PR #2, opened 2015-09-22
add run bash artifact file

drwill-ms:
Copying from devtestlabtest. Tested with a Linux VM in dev environment, and verified that custom script ran (because it put info in a file).


shichaoh:
Looks good to me!

PR #12, opened 2015-10-07
add artifacts to install several different chocolatey artifacts

mmydland:


Artifacts/windows-7zip/ChocolateyPackageInstaller.ps1
```diff
@@ -0,0 +1,262 @@
...
+#  - N/A.
+#
+
+function InstallPackages
+{
+    Param(
+        [ValidateNotNullOrEmpty()][string] $packagesList
+    )
+
+    $Separator = @(";",",")
```
drwill-ms: minor: consistent casing for local variables

ryanvog: IIRC we include this file in each artifact that uses Chocolatey.   Should we find a way to source it from a common location instead of including it in each artifact?  That would reduce any effort needed when this script changes.

Artifacts/windows-fiddler/ChocolateyPackageInstaller.ps1
```diff
@@ -0,0 +1,262 @@
...
+}
+
+##################################################################################################
+
+#
+# Description:
+#  - Installs the specified chocolatet packages on the machine.
+#
+# Parameters:
+#  - N/A.
```
drwill-ms: parameter from below?

Artifacts/windows-fiddler/ChocolateyPackageInstaller.ps1
```diff
@@ -0,0 +1,262 @@
...
+#
+# Return:
+#  - N/A.
+#
+# Notes:
+#  - N/A.
```
drwill-ms: Should you note that the license is accepted by running this script?

Artifacts/windows-7zip/ChocolateyPackageInstaller.ps1
```diff
@@ -0,0 +1,262 @@
...
+# Note: Because the $ErrorActionPreference is "Stop", this script will stop on first failure.  
+$ErrorActionPreference = "stop"
+
+###################################################################################################
+
+#
+# Custom Configurations
+#
+
+$ChocolateyPackageInstallerFolder = Join-Path $env:ALLUSERSPROFILE -ChildPath $("ChocolateyPackageInstaller-" + [System.DateTime]::Now.ToString("yyyy-MM-dd-HH-mm-ss"))
```
mithunshanbhag: It's probably best to replace $env:ALLUSERSPROFILE with $env:USERPROFILE since the latter location can be written to even by a non-admin.

Artifacts/windows-7zip/ChocolateyPackageInstaller.ps1
mithunshanbhag: Agreed.

Artifacts/windows-7zip/ChocolateyPackageInstaller.ps1
```diff
@@ -0,0 +1,262 @@
...
+        WriteLog $("No packages were specified. Exiting...")
+        return        
+    }
+
+    foreach ($package in $packages)
+    {
+        WriteLog $("Installing package: " + $package)
+
+        # install git via chocolatey
+        choco install $package --force --yes --acceptlicense --verbose | Out-Null
```
mithunshanbhag: This installs the latest available version of specified package.We should probably consider the following: 
1. Using a LKG (last known good) version. 
2. Adding an optional parameter to this script to let user explicitly specify version of the chocolatey package.

Artifacts/windows-7zip/ChocolateyPackageInstaller.ps1
mithunshanbhag: There are no formal casing rules for powershell. The one that I stick to (and the one used here) is  

1. local variables (defined within a method) and method parameters follow camel casing.
2. global variable (defined outside a method) follow pascal casing. 
3. method names follow pascal casing

drwill-ms:
Seems awfully repetitive. Seems like this will be a maintenance PITA when we want to change them all. Is there no way to abstract out what is common?

PR #21, opened 2015-11-09
Remove userName and password parameters. Remove atom artifact.

iouri-s:
Artifacts get installed using a temporary account.
Atom installation was never successful - removed it pending further investigation.


mithunshanbhag:
We're deleting the user profile after artifact installation, so we may lose any logs persisted in %USERPROFILE% by the installer (if any). 

LGTM otherwise.

PR #27, opened 2015-12-08
Add VM deletion to vms action

ryanvog:
This enables the user to delete virtual machines (and their respective environment) from the DTL CLI.  You may now include the -d or --delete switch to indicate that a vm is to be deleted.  for example:
```
dtl <credentials> vms -l mylab -n myvm -d
```

### DIFF
--- a/Scripts/linux/dtl
+++ b/Scripts/linux/dtl
@@ -95,7 +95,7 @@ class DevTestLabsMain:
 
     def __ensureAuthenticated(self):
         """Retrieves an authorization token for the principal"""
-        provider = dtlproviders.AuthorizeTokenProvider()
+        provider = dtlproviders.AuthorizeTokenProvider(self._printService)
 
         accessKey = provider.provide(self._parsedArgs)
 
@@ -144,6 +144,14 @@ class DevTestLabsMain:
 
         self._printService.verbose('Parsed argument(s):')
         self._printService.verbose(self._parsedArgs)
+
+        if self._parsedArgs.clientID is None or self._parsedArgs.secret is None or self._parsedArgs.tenant is None:
+            self._printService.error('Missing required service principal credentials')
+            sys.exit(1)
+
+        if self._parsedArgs.subscription is None:
+            self._printService.error('Missing required subscription')
+            sys.exit(1)
 
         return
 
--- a/Scripts/linux/dtlcommands.py
+++ b/Scripts/linux/dtlcommands.py
@@ -561,7 +561,7 @@ class VirtualMachinesAction:
                                dest='vmid',
                                help='The id of the virtual machine to retrieve.',
                                required=False)
-        argParser.add_argument('--d', '--delete',
+        argParser.add_argument('-d', '--delete',
                                dest='deletevm',
                                action='store_true',
                                help='Deletes the virtual machine resource from the lab.',
--- a/Scripts/linux/dtlcommands.py
+++ b/Scripts/linux/dtlcommands.py
@@ -160,7 +160,7 @@ class AuthorizeCommandAction:
         Returns:
             None
         """
-        return 'Authorizes access to Azure resources'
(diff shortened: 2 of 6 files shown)


Scripts/linux/dtl
```diff
@@ -145,6 +145,14 @@ class DevTestLabsMain:
         self._printService.verbose('Parsed argument(s):')
         self._printService.verbose(self._parsedArgs)
 
+        if self._parsedArgs.clientID is None or self._parsedArgs.secret is None or self._parsedArgs.tenant is None:
+            self._printService.error('Missing required service principal credentials')
```
ericame: It would be clearer here if you listed which args were required as none of them are actually called service principal

Scripts/linux/dtlcommands.py
```diff
@@ -585,25 +590,29 @@ def invoke(self, settings):
         printService = dtlprint.PrintService(settings.quiet, settings.verbose)
         labSvc = dtllabservice.LabService(settings, printService)
 
-        if settings.name is not None:
-            vms = labSvc.getVirtualMachine(settings.subscription, name=settings.name)
-        elif settings.vmid is not None:
-            vms = labSvc.getVirtualMachine(settings.subscription, vmId=settings.vmid)
+        if settings.deletevm and (
+                        settings.name is not None or settings.vmid is not None) and settings.labname is not None:
```
ericame: I know this was added earlier, but consider calling it vmname instead of name in case we need to use the name of something else later - while this is context specific it will just be a little more intuitive

Scripts/linux/dtllabservice.py
```diff
@@ -332,55 +332,128 @@ def getVirtualMachine(self, subscriptionId, vmId=None, name=None):
         Returns:
             The virtual machine that matches the provided filter, or None.
 
-        Example retun data:
+        Example return data:
```
ericame: Is there any way to get the service to give you the expected output so we don't have to update this every time they change?

Scripts/linux/dtllabservice.py
```diff
@@ -332,55 +332,128 @@ def getVirtualMachine(self, subscriptionId, vmId=None, name=None):
...
+        vms = []
+
+        for environment in environments:
+            for vm in environment['properties']['vms']:
+                if vm[fieldName].lower() == fieldValue.lower():
+                    vm['environmentId'] = environment['id']
+                    vms.append(vm)
+                    continue
+
+        return vms
```
ericame: return None if vm in environment is none?

Scripts/linux/dtllabservice.py
```diff
@@ -332,55 +332,128 @@ def getVirtualMachine(self, subscriptionId, vmId=None, name=None):
...
+            name (string)           - The name of tha virtual machine.  If there are more than one virtual machine with
+                                      the same name in the specified lab, you must use the vmId parameter instead.
+            vmId (string)           - The resource ID of the virtual machine.
+        Returns:
+            0 if successful, 1 otherwise.
+        """
+
+        vms = self.getVirtualMachine(subscriptionId, name, vmId)
+
+        if vms is None or len(vms) == 0:
```
ericame: Consider moving the retrieve required VM logic to its own function so you're not repeating the validation logic in every function that needs a vm

Scripts/linux/dtllabservice.py
```diff
@@ -332,55 +332,128 @@ def getVirtualMachine(self, subscriptionId, vmId=None, name=None):
...
+                    name))
+            return 1
+
+        vm = vms[0]
+
+        self._printService.info('Deleting virtual machine: ' + vm['name'])
+
+        environmentId = vm['environmentId']
+
+        url = environmentId + '?api-version={0}'.format(self._apiVersion)
```
ericame: This might be a tad more readable:
`url = '{0}?api-version{1}'.format(environmentId, self._apiVersion)`

ericame:
LGTM

PR #56, opened 2016-03-03
Jaymar/new community update1

jaymarf:
Updating the Arm Template and PowerShell commandlets to allow sysprep of a VM when Creating a Custom Image


mithunshanbhag:
Confirmed with @jaymarf that this PR can be closed without merging.

PR #57, opened 2016-03-03
Jaymarf patch 1

jaymarf:


mithunshanbhag:
Confirmed with @jaymarf that this PR can be closed without merging.

PR #60, opened 2016-03-08
remove the powershell artifact

mmydland:
Remove the public install powershell artifact which is not working correctly.


ryanvog:
LGTM

PR #65, opened 2016-03-14
add slack and sublime text artficacts

ins4n333:


azurecla:
Hi __@ins4n333__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        In order for us to evaluate and accept your PR, we ask that you sign a contribution license agreement. It's all electronic and will take just minutes. I promise there's no faxing. https://cla.azure.com.
    </span>

TTYL, AZPRBOT;

Artifacts/windows-sublime-text/Artifactfile.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2015-01-01/dtlArtifacts.json",
+  "title": "Sublime Text",
+  "description": "Installs SublimeText using the Chocolatey package manager",
+  "tags": [
+    "Windows",
+    "Fiddler"
```
drwill-ms: Did you mean Sublime?

ins4n333: Oops) I will fix it

ins4n333:
done)

Artifacts/windows-slack/Artifactfile.json
```diff
@@ -0,0 +1,26 @@
...
+  "title": "Slack",
+  "description": "Installs Slack using the Chocolatey package manager",
+  "tags": [
+    "Windows",
+    "Slack"
+  ],
+  "iconUri": "https://chocolatey.org/content/packageimages/slack.1.0.0.png",
+  "targetOsType": "Windows",
+   "parameters": {
+        "installUsername": {
```
mithunshanbhag: I see that this artifact requires the user to pass in a username and password (for installing slack under a specific user's credentials). Would it be possible to eliminate the need for these parameters?

The general pattern that most artifacts follow is creation of a temporary admin account to install the chocolatey package (Example: https://github.com/Azure/azure-devtestlab/tree/master/Artifacts/windows-7zip). This way, the user is not required to pass in a user name and password.

Will that work for this artifact?

ins4n333: The key problem is that Slack (with this artifact) does not install synchronous . Slack will be installed on the machine only after the user (which specified in artifcat parameters), connect via RDP to this machine. It is a feature of the msi installer that is used in the choco package. Therefore, in this artifact I cannot use temporary user.

ins4n333: Initially, I just wanted to use  chocolatey package "slack". But  this package uses a custom installer which does not work in silent mode.. Installation of the artifact hung during the creation of the machine. But I found a standart msi installer And I create a new package (on myget feed [not chocolatey.org]).

azurecla:
__@ins4n333__, Thanks for signing the contribution license agreement so quickly! Actual humans will now validate the agreement and then evaluate the PR.
<br />Thanks, AZPRBOT;

Artifacts/windows-slack/Artifactfile.json
ramankumarlive: @drwill-ms and @mithunshanbhag, can you please recommend any changes that need to be made to the artifact? If there are no recommendations, can we merge the artifact with master?

drwill-ms: Makes sense. So what credentials are being supplied? The username and password of the VM user which is deployed with the VM, or are they Slack-specific? I think that ambiguity should be cleared up in the parameter descriptions.

ins4n333: @drwill-ms I originally wrote this information in the README.md :)
{userName} and {password} - This is credentials of the user for which you want to install the package. After executing this command, you must  make a logout. 

Ок. I add similar tips to the parameter descriptions.

drwill-ms: I appreciate that, but users won't see the readme or a link to it when they use the Azure portal UI to create VMs and add artifacts, so having it clearer in the parameter name and description will make a big difference.

Still, I think the phrase "credentials of user for hich you want to install the package" isn't necessarily clear between local VM account and Slack account. I think you mean the former, right?

PR #83, opened 2016-05-25
Jraj/demo lab

joerage:


mmydland:
seems like this is so old that it can't be pertinent anymore.  If you still want to merge it, please submit a new pull request.

PR #128, opened 2016-08-30
Added artifact to install Powershell 3.

joerage:


Artifacts/windows-powershell3/install.cmd
```diff
@@ -0,0 +1,14 @@
...
+
+reg Query "HKLM\Hardware\Description\System\CentralProcessor\0" | find /i "x86" > NUL && set OSVERSION=32BIT || set OSVERSION=64BIT
+
+if %OSVERSION%==32BIT (
+  call Windows6.1-KB2506143-x86.msu /quiet /norestart
+) else (
+  call Windows6.1-KB2506143-x64.msu /quiet /norestart
+)
+
+REM ErrorLevel means "The requested operation is successful. Changes will not be effective until the system is rebooted."
```
drwill-ms: ErrorLevel 3010 means...

drwill-ms:
Looks good

Artifacts/windows-powershell3/install.cmd
leovms: "take effect" instead of "be effective"

Artifacts/windows-powershell3/install.cmd
```diff
@@ -0,0 +1,14 @@
+@echo OFF
```
leovms: Have you considered doing this in PS instead of a batch file? You would have more control. Also, is there a way to download the .MSU files instead of making them part of the artifact? That way you only download what is needed and we can also ensure it is the latest.

joerage: I thought about using PS, but batch file seems simpler for the task. Although I did not try, one of the failing condition of this MSU is if you have a powershell window opened. So it might have interfered...
As for downloading, it would have made the script more complex. This MSU is pretty old, I don't think there is much chance of being updated. And I believe it would be a different file name anyway.

Artifacts/windows-powershell3/install.cmd
joerage: done.

PR #145, opened 2016-10-18
Fix chocolatey artifact package installation scripts

leovms:
- Added better error handling for startChocolatey.ps1
- Updated all chocolatey artifact definitions to correctly call PS script to handle exit codes.
- Implemented fix to work in Windows 2016.


drwill-ms:
LGTM

PR #147, opened 2016-10-19
Added missing quotes in definition of PackageList

leovms:


leovms:
Too many conflicts. I need to rebase separately. Applied fixes directly.

PR #169, opened 2016-11-17
Fix errors in the clone-git artifact

iouri-s:
Start by initializing folders
Catch errors caused by logging errors
Do not use single quotes on Windows command line


drwill-ms:
Looks good to me

PR #175, opened 2016-11-28
Add pattern validation to iconUri property, update descriptions

drwill-ms:


schemas/2015-01-01/dtlArtifacts.json
```diff
@@ -14,14 +14,16 @@
...
         "description": {
             "type": "string",
-            "description": "The description of your artifact."
+            "description": "The artifact's description."
         },
         "iconUri": {
-            "type": "string"
+            "type": "string",
+            "description": "An icon to identify and distinguish the artifact in a list. URL referencing the icon must use SSL.",
+            "pattern": "(^https:\/\/.*)|(^$)"
```
leovms: You could try to simplify to `https:\/\/.*|^$` to avoid creating unnecessary groups.

PR #176, opened 2016-11-29
Update to latest artifact schema

drwill-ms:


Artifacts/linux-install-nodejs/linux_install_nodejs_v1.sh
```diff
@@ -7,7 +7,7 @@ if [ -f /usr/bin/apt ] ; then
 
     apt-get -y update
     
-    curl --silent --location https://deb.nodesource.com/setup_0.12 | bash -
+    curl --silent --location https://deb.nodesource.com/setup_4.x | bash -
```
drwill-ms: This artifact was hanging on Ubuntu 16.10 and console output said the source was being deprecated. It referred me to this version as the preferred one. This works on Ubuntu 16.10 and LTS.

ericame:
lgtm

PR #179, opened 2016-12-06
- Ensure not to run the install script is docker is already installed.

joerage:
- Fixed a bug when parameters where not passed in, artifact installation would hang forever. Forced having an explicit container name to avoid parameters confusion.


drwill-ms:
Looks good

PR #182, opened 2016-12-13
Fix windows-download-sas-blob

leovms:
- Updated syntax used to execute the powershell command.
- Restructured the logic in the PS1 to ensure errors correctly bubble up.


drwill-ms:
looks good

PR #183, opened 2016-12-14
Update VSCode artifact to support VMs with older powershell versions

workleodig:
Also added the !runcode parameter to the install so that it does not start Code.exe when the install completes


azurecla:
Hi __@workleodig__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        This seems like a small (but important) contribution, so no contribution license agreement is required at this point. Real humans will now evaluate your PR.
    </span>

TTYL, AZPRBOT;

leovms:
Also, we will be moving more toward leveraging Chocolatey. So, this artifact will get updated to use the package **visualstudiocode**.

PR #186, opened 2016-12-20
Fix for Azure PowerShell artifact shows succeeded but actually failed

leovms:
- Rebased startChocolatey.ps1 and ChocolateyPackageInstaller.ps1 to match all other chocolatey based artifacts.
- Updated artifact to ensure errors bubble up correctly.
- Converted artifact to use chocolatey package "windowsazurepowershell".
- Removed the need to install WebPICmd.


drwill-ms:
looks good

PR #193, opened 2017-02-18
Fix for validation error on driveLetter

reddobowen:
DriveLetter needs to be forced into a String, otherwise the script will Error: Cannot validate argument on parameter 'driveLetter'.


azurecla:
Hi __@reddobowen__, I'm your friendly neighborhood Azure Pull Request Bot (You can call me AZPRBOT). Thanks for your contribution!
    <span>
        This seems like a small (but important) contribution, so no contribution license agreement is required at this point. Real humans will now evaluate your PR.
    </span>

TTYL, AZPRBOT;

PR #194, opened 2017-02-20
Adding postDeployActions

leovms:
Adding support for post deploy actions to help control VM restart/deallocate outside of the artifact implementation. Because the option is not a breaking change, we update the current definition and still remain backward compatible.


joerage:
All good.

PR #195, opened 2017-02-23
Moving poweroff logic out of artifact

leovms:
Moving poweroff logic out of the artifact and leveraging DTL support for shutting down the VM after deprovisioning.

### DIFF
--- a/Artifacts/linux-deprovision/depro.sh
+++ b/Artifacts/linux-deprovision/depro.sh
@@ -2,7 +2,7 @@
 
 main () {
     { # try
-        echo 'waagent -force -deprovision+user && poweroff' | at now + 1 minute > /dev/null 2>&1
+        echo 'waagent -force -deprovision+user' | at now + 1 minute > /dev/null 2>&1
         exit 0
     } || { # catch
         exit 1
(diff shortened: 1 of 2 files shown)


Artifacts/linux-deprovision/depro.sh
```diff
@@ -2,7 +2,7 @@
 
 main () {
     { # try
-        echo 'waagent -force -deprovision+user && poweroff' | at now + 1 minute > /dev/null 2>&1
+        echo 'waagent -force -deprovision+user' | at now + 1 minute > /dev/null 2>&1
```
joerage: Do you still need a delay for deprovisioning?

leovms: Yes, I do. The `waagent` command is scheduled to run 1 minute later, using the `at` command. If we don't put a delay, the artifact will return right away and a shutdown will start. However, the `waagent` may not have fully completed its work. So, adding a delay for the shutdown will ensure that the command completes.

joerage: But you already have a 2 minutes delay in the postDeployActions. Do you need both?

PR #199, opened 2017-02-27
Fix issues with the checksum for google artifact

mmydland:
ignoring the checksum in the googlechrome package from chocolatey.  The msi is downloaded from a google site and is unversioned so this artifact is going to break from time to time without this change.

### DIFF
--- a/Artifacts/windows-chrome/Artifactfile.json
+++ b/Artifacts/windows-chrome/Artifactfile.json
@@ -10,6 +10,6 @@
   "iconUri": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/Artifacts/windows-chrome/Google_Chrome_icon.png",
   "targetOsType": "Windows",
   "runCommand": {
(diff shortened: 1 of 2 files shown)


Artifacts/windows-chrome/startChocolatey.ps1
```diff
@@ -223,7 +223,7 @@ function Invoke-ChocolateyPackageInstaller
     $oldPolicyValue = Set-LocalAccountTokenFilterPolicy
     try
     {
-        Invoke-Command -ComputerName $env:COMPUTERNAME -Credential $credential -FilePath $command -ArgumentList $PackageList
+        Invoke-Command -ComputerName $env:COMPUTERNAME -Credential $credential -FilePath $command -ArgumentList $PackageList --ignorechecksum
```
drwill-ms: How about a comment about why we ignore the checksum?

drwill-ms:
Aside from comment about ignoring the checksum, looks good.

PR #212, opened 2017-04-06
Add SSMS artifact

azabbasi:
I have added the SSMS (sql server management studio) artifact to our public artifact


leovms:
Use this [ChocolateyPackageInstaller.sp1](https://github.com/Azure/azure-devtestlab/blob/master/Artifacts/windows-azurepowershell/ChocolateyPackageInstaller.ps1) version instead, as it also handles restart requirements without failing the artifact.

Issue #218, opened 2017-04-19
The package was not found with the source(s) listed (invalid package name).


andremarques023:
Microsoft .NET Framework 4.5 artifact fails when installing. I've found that the package name defined on Artifactfile.json is actually wrong.

drwill-ms:
Thanks for submitting a fix

PR #219, opened 2017-04-19
Fixed chocolatey package name.

andremarques023:
The package name defined on Artifactfile.json is actually wrong. It should be dotnet4.5 otherwise it will fail, since the package will not be found on source (https://chocolatey.org/api/v2/).


andremarques023:
Related to #218.

drwill-ms:
Thank you

PR #233, opened 2017-05-22
Added support for script arguments for the windows-vsts-download-and-run-script artifact

gregpakes:
This enabled support for script arguments for the script to run

### DIFF
--- a/Artifacts/windows-vsts-download-and-run-script/Artifactfile.json
+++ b/Artifacts/windows-vsts-download-and-run-script/Artifactfile.json
@@ -30,7 +30,8 @@
             "type": "string",
             "displayName": "Script Arguments",
             "description": "Optional arguments for the script, eg '-username John -server localhost'",
-            "allowEmpty": true
+            "allowEmpty": true,
+            "defaultValue": ""            
         }
     },
     "runCommand": {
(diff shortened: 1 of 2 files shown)


drwill-ms:
Thanks Greg. Does the artifact still work if parameters are not included?

gregpakes:
@drwill-ms 

I see two potential issues here.

1. Existing ARM templates won't have this parameter included.
2. Does it work if the parameter is empty?

**Answers**

1. If I submit an ARM template deployment without these parameters, I get:

>Unable to generate valid template for the artifact - Download VSTS Drop and Run Script with error(s) - Deployment template validation failed: 'The value for the template parameter 'scriptArguments' at line '25' and column '29' is not provided. Please see https://aka.ms/arm-deploy/#parameter-file for usage details.'.

Is there a way to make these optional, even in the ARM template?  This must be a common issue when updating the artifacts?

2. This works as expected.  The parameters are ignored.

drwill-ms:
I've run into that myself. PowerShell doesn't like providing a parameter but not providing a value for it, and our artifactFile.json format doesn't support conditionally adding parameters. The only way I've been able to get around this is if parameters are passed in order. However, that is generally an undesirable way to do parameters these days, especially in PowerShell.

What if you provide a default value in the artifact? That way at least something gets passed to the script?

gregpakes:
How would I do that?  The docs don't suggest there is an option for default value?

https://docs.microsoft.com/en-us/azure/devtest-lab/devtest-lab-artifact-author

gregpakes:
**Update**  I've looked at some other artifacts and found the defaultValue property.  Testing now.

gregpakes:
@drwill-ms 

This works as expected now!

thanks!

drwill-ms:
Great! We'll run a few quick tests on it and then merge the changes. Thanks!

PR #254, opened 2017-06-22
Updating artifact structure and logic

leovms:
- Handling more edge scenarios where the VS installer returns error code other than 0.
- Ensuring installation can succeed consistently.

### DIFF
--- a/Artifacts/windows-ServiceFabricSDK/InstallServiceFabricSDKAndTools.ps1
+++ b/Artifacts/windows-ServiceFabricSDK/InstallServiceFabricSDKAndTools.ps1
@@ -278,6 +278,7 @@ function Enable-ServiceFabricTools
 {
     [CmdletBinding()]
     param(
+        [int] $VSVersionNumber
     )
 
     $vsBootstrapperExe = Join-Path $env:Temp "vsbootstrap.exe"
@@ -308,7 +309,7 @@ function Enable-ServiceFabricTools
 
     # Add Service Fabric component to all Visual Studio instances.
     $vsInstances | % {
-        if ($_.InstallationVersion.Major -eq 15)
+        if ($_.InstallationVersion.Major -eq $VSVersionNumber)
         {
             # We must do an update to restore the channel used for modifying the Visual Studio Instance.
             Write-Host 'Updating Visual Studio instance'
@@ -317,7 +318,6 @@ function Enable-ServiceFabricTools
             # Modify the Visual Studio instance with Service Fabric SDK components.
             Write-Host "Enabling Service Fabric Tools component for installation $($_.InstallationPath)"
             Invoke-Process -FileName $vsInstallExe -Arguments "modify --installPath `"$($_.InstallationPath)`" --add Microsoft.VisualStudio.Workload.Azure --add Microsoft.VisualStudio.Component.Azure.ServiceFabric.Tools --quiet --norestart --wait" -ValidExitCodes 1
-            break;
         }
     }
 }
@@ -362,7 +362,7 @@ try
     if ($vsVersionNumber -ge 15)
     {
         Write-Host 'Enabling Service Fabric Tools'
-        Enable-ServiceFabricTools
+        Enable-ServiceFabricTools -VSVersionNumber $vsVersionNumber
     }
 }
 finally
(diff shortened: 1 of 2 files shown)


Artifacts/windows-ServiceFabricSDK/InstallServiceFabricSDKAndTools.ps1
```diff
@@ -1,136 +1,375 @@
...
+    if (-not (Test-Path $vsBootstrapperExe))
+    {
+        throw "Visual Studio bootstrapper was not successfully downloaded to $vsBootstrapperExe"
     }
-    StartAndWaitForProcess $vsBootstrapperExe "--quiet --update" @(0)
 
-    #Get path to vs_installer.exe
-    $vsInstallerExePath = (Get-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\{6F320B93-EE3C-4826-85E0-ADF79F8D4C61}' -Name 'InstallLocation').InstallLocation.Trim('"')
+    Write-Host 'Getting Visual Studio Installer path'
+    $vsInstallerExePath = (Get-ItemProperty -Path "Registry::HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\{6F320B93-EE3C-4826-85E0-ADF79F8D4C61}" -Name 'InstallLocation').InstallLocation.Trim('"')
```
drwill-ms: mixed apostrophes and quotation marks for parameters

Artifacts/windows-ServiceFabricSDK/InstallServiceFabricSDKAndTools.ps1
```diff
@@ -1,136 +1,375 @@
...
+        if ($_.InstallationVersion.Major -eq 15)
+        {
+            # We must do an update to restore the channel used for modifying the Visual Studio Instance.
+            Write-Host 'Updating Visual Studio instance'
+            Invoke-Process -FileName $vsBootstrapperExe -Arguments "update --installPath `"$($_.InstallationPath)`" --quiet --wait" -ValidExitCodes 1
+         
+            # Modify the Visual Studio instance with Service Fabric SDK components.
+            Write-Host "Enabling Service Fabric Tools component for installation $($_.InstallationPath)"
+            Invoke-Process -FileName $vsInstallExe -Arguments "modify --installPath `"$($_.InstallationPath)`" --add Microsoft.VisualStudio.Workload.Azure --add Microsoft.VisualStudio.Component.Azure.ServiceFabric.Tools --quiet --norestart --wait" -ValidExitCodes 1
+            break;
```
EMaher: We should either update all installed instances of that VS version or allow use to specify which instance of VS they want to modify. Original version of artifact affected all instances.

Artifacts/windows-ServiceFabricSDK/InstallServiceFabricSDKAndTools.ps1
leovms: As a general best practice I stick to single quotes for all literal strings. I use double quotes for strings that I want expanded (`"$myVar"`), or where I want to use escape characters (``"Break this over `n two lines"``).

In the case of the `Get-ItemProperty` call, we need to use double quotes so that the `"Registry::HKLM"` gets expanded correctly. Otherwise, the call will fail.

Artifacts/windows-ServiceFabricSDK/InstallServiceFabricSDKAndTools.ps1
leovms: Corrected.

Artifacts/windows-ServiceFabricSDK/InstallServiceFabricSDKAndTools.ps1
```diff
@@ -1,136 +1,375 @@
...
-}
\ No newline at end of file
+    Write-Host "Visual Studio Installer is at $vsInstallExe"
+
+    # Get VS installation information.
+    $vsInstances = Get-VSSetupInstances
+
+    # Add Service Fabric component to all Visual Studio instances.
+    $vsInstances | % {
+        if ($_.InstallationVersion.Major -eq 15)
```
leovms: Good point, I will pass the version number as a parameter, as we should always refrain from using "global" variables, where possible.

PR #262, opened 2017-07-14
new artifact to install JHipster on Windows

markusheiliger:


Artifacts/windows-jhipster/Artifactfile.ps1
```diff
@@ -0,0 +1,294 @@
+﻿<##################################################################################################
```
mmydland: May want to keep the same name as the other startChocolatey script files so that we catch them when we need changes

PR #268, opened 2017-07-27
Allow agentName to use the computer's name by default

zumwald:
At least in scenarios within my team, we want to be able to spin up a set of machines (say, 15 or even 50) from a single action (see image below). This isn't currently possible to successfully do this and apply the VSTS Build Agent artifact in order to provision each of the machines in a VSTS agent pool.

This pull request enables that scenario by allowing the agent name to be empty (minor `Artifactfile.json` changes) and then removes the automatic `param` validation and instead sets `$agentName` to `%COMPUTERNAME%` if the parameter wasn't supplied by the user.

image

### DIFF
--- a/Artifacts/windows-vsts-build-agent/Artifactfile.json
+++ b/Artifacts/windows-vsts-build-agent/Artifactfile.json
@@ -25,7 +25,8 @@
     "agentName": {
       "type": "string",
       "displayName": "Agent Name",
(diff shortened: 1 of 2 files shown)


zumwald:
@leovms thoughts or concerns?

leovms:
Looks good. Thank you for your contribution.

PR #272, opened 2017-08-10
No op artifact

abhisshams:
Adding a no op artifact for windows,


Artifacts/windows-noop/Artifactfile.json
```diff
@@ -0,0 +1,13 @@
+{
+    "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2016-11-28/dtlArtifacts.json",
+    "title": "No-Op",
+    "publisher": "Microsoft",
+    "description": "Artifact internally used to work around Custom Script extension upgrade limitations on sysprepped images.",
```
ajay-arora: Confirm from PMs/Mark if we really want this artifact to have this description. Because I am not sure if we want to.

PR #284, opened 2017-09-13
Add support for SLES zypper to Linux-deprovision

michaelcohenmicrosoft:


michaelcohenmicrosoft:
@leovms Adding in linux-deprovision support for zypper package manager on SLES

Issue #289, opened 2017-10-05
Encoding Issues on CopyVHDFromVMToLab.ps1


akalcik:
There must be some encoding issues with the [CopyVHDFromVMToLab](https://github.com/Azure/azure-devtestlab/blob/master/Scripts/CopyVHDFromVMToLab.ps1) as I wasn't able to run this script on my Windows machine. When I run the script I get the error message:

```
+ }
+ ~
Unexpected token '}' in expression or statement.
    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : UnexpectedToken
```

Now when I open the script in the PowerShell ISE there are some encoding issues which leads to error before.

Line 192
```powershell
$destContext = New-AzureStorageContext â€“StorageAccountName $CopyContext.LabStorageAccountName -StorageAccountKey $CopyContext.LabStorageAccountKey
```

Line 200
```powershell
$srcContext = New-AzureStorageContext â€“StorageAccountName $CopyContext.VMStorageAccountName -StorageAccountKey $CopyContext.VMStorageAccountKey
```

When I replace the ```â€``` with ```-``` it corrects the parsing issue.

leovms:
Thank you. I have gone ahead and made the corrections.

PR #290, opened 2017-10-05
Add SignatureExpire parameter and make some parameters mandatory

akalcik:
I fixed issues #289 and also introduced some improvements.  I added  ```SignatureExpire``` parameter because sometimes the default time in 3600 seconds isn't enough which leads to copy operation will be canceled. I also made some parameters as mandatory because it doesn't have a sense to run the script with some default values.


leovms:
Thank you for your contribution.

PR #296, opened 2017-10-18
Fix broken linux artifacts 

ajay-arora:
Artifacts were broken because set -e was set too early in the script. As some of the commands in the script are expected to fail.

### DIFF
--- a/Artifacts/linux-install-mongodb/linux_install_mongodb_v1.sh
+++ b/Artifacts/linux-install-mongodb/linux_install_mongodb_v1.sh
@@ -1,9 +1,11 @@
 #!/bin/bash
 
-set -e
-
 isApt=`command -v apt-get`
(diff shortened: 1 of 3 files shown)


Artifacts/linux-install-mongodb/linux_install_mongodb_v1.sh
```diff
@@ -1,10 +1,14 @@
 #!/bin/bash
 
-set -e
+echo "Installing MongoDb"
 
 isApt=`command -v apt-get`
 isYum=`command -v yum`
 
+# The above commands may error on some flavors depending on which installation mechanism is available.
```
drwill-ms: I'd say: Some of the previous commands will fail with an exit code other than zero (intentionally), so we do not set error handling to stop (set e) until after they've run.

PR #299, opened 2017-10-30
Create artifacts to install VS2015 and 2017

workleodig:


Artifacts/windows-visualstudio2015/artifact.ps1
```diff
@@ -0,0 +1,93 @@
...
+
+function DownloadToFilePath ($downloadUrl, $targetFile)
+{
+    Write-Output ("Downloading installation files from URL: $downloadUrl to $targetFile")
+    $targetFolder = Split-Path $targetFile
+
+    if((Test-Path -path $targetFile))
+    {
+        Write-Output "Deleting old target file $targetFile"
+        Remove-Item $targetFile | Out-Null
```
leovms: Consider using `-Force` so the call won't fail, especially since you are hiding the output.

Artifacts/windows-visualstudio2015/artifact.ps1
```diff
@@ -0,0 +1,93 @@
...
+}
+elseif($sku -eq 'Enterprise') {
+    $downloadUrl = 'http://go.microsoft.com/fwlink/?LinkId=615437'
+    $downloadFile = "vs_enterprise.exe"
+}
+else {
+    Write-Error "SKU is not recognized - allowed values are Professional and Enterprise. Specified value: $sku"
+}
+
+#VS Install log file
```
leovms: Remove. Seems redundant.

Artifacts/windows-visualstudio2015/artifact.ps1
```diff
@@ -0,0 +1,93 @@
...
+
+$argumentList = "$installerArgs /Quiet /NoRestart /Log $vsLog"
+Write-Output "Running install with the following arguments: $argumentList"
+
+#Run the install and wait for completion.
+$retCode = Start-Process -FilePath $localFile -ArgumentList $argumentList -Wait -PassThru
+
+if ($retCode.ExitCode -ne 0)
+{
+	Write-Output ("Visual Studio installed with reboot: " + $retCode.ExitCode.ToString())
```
leovms: Spacing.

Artifacts/windows-visualstudio2015/artifact.ps1
```diff
@@ -0,0 +1,93 @@
...
+    Restart-Computer -Force
```
leovms: I strongly recommend you re-write this logic to leverage the `postDeployActions` in the JSON definition instead, to add reboot/restart logic.

Artifacts/windows-visualstudio2015/artifact.ps1
```diff
@@ -0,0 +1,93 @@
...
+
+Write-Output "Downloaded files: $localFile"
+
+$argumentList = "$installerArgs /Quiet /NoRestart /Log $vsLog"
+Write-Output "Running install with the following arguments: $argumentList"
+
+#Run the install and wait for completion.
+$retCode = Start-Process -FilePath $localFile -ArgumentList $argumentList -Wait -PassThru
+
+if ($retCode.ExitCode -ne 0)
```
leovms: Also, consider checking for exit code **3010** explicitly, as it indicates the VM needs rebooting. But you don't want to fail your artifact because of it, especially if you are rebooting anyway. See artifact [Install Chocolatey (ps1)](https://github.com/Azure/azure-devtestlab/blob/master/Artifacts/windows-chocolatey/install-choco-package.ps1) for an example, including a proper message to output in the logs to that effect.

Artifacts/windows-visualstudio2015/Artifactfile.json
```diff
@@ -0,0 +1,32 @@
...
+    "installerArgs": {
+      "type": "string",
+      "defaultValue": "",
+      "allowEmpty": true,
+      "displayName": "Optional args passed to VS installer",
+      "description": "Additional arguments to pass when installing VS."
+    }
+  },
+  "runCommand": {
+    "commandToExecute": "[concat('powershell.exe -ExecutionPolicy bypass \"& ./artifact-main.ps1', ' -sku ', parameters('sku'), ' -installerArgs ', '''', parameters('installerArgs'), '''', '\"')]"
```
leovms: Any chance the `sku` could contains spaces? If so, you may want to wrap it around quotes like you do for `installerArgs`.

workleodig: I'll wrap it in quotes. The artifact param has an allowedValues set and nothing contains spaces there today, but it's safer in quotes.

Artifacts/windows-visualstudio2015/artifact.ps1
workleodig: good call

Artifacts/windows-visualstudio2015/artifact.ps1
workleodig: Exit code 0 and 3010 will be considered Success. All others are logged as failure.

Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -0,0 +1,122 @@
...
+{
+    if($version -eq '2017')
+    {
+        $targetLogs = 'c:\VS2017Logs'
+        New-Item -ItemType Directory -Force -Path $targetLogs | Out-Null
+        Write-Output ('Temp location is ' + $env:TEMP)
+        Copy-Item -path $env:TEMP\dd* -Destination $targetLogs
+    }
+
+    Write-Error ("Product installation of $localFile failed with exit code: " + $retCode.ExitCode.ToString())
```
leovms: The following works too:
```
Write-Error ("Product installation of $localFile failed with exit code: $($retCode.ExitCode)")
```

Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -0,0 +1,122 @@
...
+    Write-Output ("Downloading installation files from URL: $downloadUrl to $targetFile")
+    $targetFolder = Split-Path $targetFile
+
+    if((Test-Path -path $targetFile))
+    {
+        Write-Output "Deleting old target file $targetFile"
+        Remove-Item $targetFile -Force | Out-Null
+    }
+
+    if((Test-Path -path $targetFolder) -eq $false)
```
leovms: The following seems more natural, but very subjective, of course:
```
if (-not (Test-Path -path $targetFolder))
```

Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -0,0 +1,122 @@
...
+    {
+        $downloadAttempts++
+
+        try
+        {
+            $WebClient = New-Object System.Net.WebClient
+            $WebClient.DownloadFile($downloadUrl,$targetFile)
+            break
+        }
+        catch [Exception]
```
leovms: Not sure if you need to specify `[Exception]`. `$` should still point to the exception that was raised.

Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -0,0 +1,122 @@
...
+        {
+            Write-Output "Caught exception during download..."
+            if ($_.Exception.InnerException){
+                $exceptionMessage = $_.InnerException.Message
+                Write-Output "InnerException: $exceptionMessage"
+            }
+            else {
+                $exceptionMessage = $_.Message
```
leovms: Are you sure this is not supposed to be `$_.Exception.Message`?

Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -0,0 +1,122 @@
...
+            $WebClient = New-Object System.Net.WebClient
+            $WebClient.DownloadFile($downloadUrl,$targetFile)
+            break
+        }
+        catch [Exception]
+        {
+            Write-Output "Caught exception during download..."
+            if ($_.Exception.InnerException){
+                $exceptionMessage = $_.InnerException.Message
+                Write-Output "InnerException: $exceptionMessage"
```
leovms: You could collapse these 2 lines in to
```
Write-Output "InnerException: $($_.InnerException.Message)"
```

Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -0,0 +1,122 @@
...
+            }
+            else {
+                $exceptionMessage = $_.Message
+                Write-Output "Exception: $exceptionMessage"
```
leovms: Same here, you can collapse into
```
Write-Output "Exception: $($_.Exception.Message)"
```

Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -0,0 +1,122 @@
...
+        $downloadUrl = 'http://go.microsoft.com/fwlink/?LinkId=615435'
+    }
+    elseif($sku -eq 'Enterprise') {
+        $downloadUrl = 'http://go.microsoft.com/fwlink/?LinkId=615437'
+    }
+}
+elseif ($version -eq '2017')
+{
+    if($sku -eq 'Professional') {
+        $argumentList = "--quiet --norestart --wait --add Microsoft.VisualStudio.Workload.Azure --add Microsoft.VisualStudio.Workload.Data --add Microsoft.VisualStudio.Workload.ManagedDesktop --add Microsoft.VisualStudio.Workload.NativeDesktop --add Microsoft.VisualStudio.Workload.NetCoreTools --add Microsoft.VisualStudio.Workload.NetWeb --includeRecommended"
```
leovms: Not that it will be a problem now, but if you want to make sure the `$argumentList` doesn't get expanded, you may want to surround it by single quotes instead of double quotes.

Artifacts/windows-visualstudio/artifact.ps1
workleodig: Either way is fine with me.

Artifacts/windows-visualstudio/artifact.ps1
workleodig: Good point

Artifacts/windows-visualstudio/artifact.ps1
workleodig: Done here and in the else block.

Artifacts/windows-visualstudio/artifact.ps1
workleodig: You're right.

Artifacts/windows-visualstudio/artifact.ps1
workleodig: done

Artifacts/windows-visualstudio/artifact.ps1
workleodig: Sure. I can use single quotes here

PR #301, opened 2017-11-01
Passing VM OS information as object parameter

leovms:
Updated approach used to set the VM OS information when preparing the request to create the custom image.

Before, we used the osType to distinguish in the RP and choose either the linuxOsState or windowsOsState. We have since moved to simply check if the OS State object is present and default to the Windows OS State check first. So, if both objects are passed, which is what we did before, we take Windows even though the OS Type was set to Linux. Therefore, I have change the logic in the task to ensure we only pass the corresponding OS State based on the OS Type. We do the same thing from the Azure Portal logic. I matched that logic.

### DIFF
--- a/VstsTasks/AzureDtlCreateCustomImage/new-azuredtl-customimage.json
+++ b/VstsTasks/AzureDtlCreateCustomImage/new-azuredtl-customimage.json
@@ -8,34 +8,8 @@
         "labName": {
             "type": "string"
         },
-        "sourceLabVmId": {
-            "type": "string"
-        },
-        "osType": {
-            "type": "string",
-            "allowedValues": [
-                "Linux",
-                "Windows"
-            ],
-            "defaultValue": "Windows"
-        },
-        "linuxOsState": {
-            "type": "string",
-            "allowedValues": [
-                "NonDeprovisioned",
-                "DeprovisionRequested",
-                "DeprovisionApplied"
-            ],
-            "defaultValue": "NonDeprovisioned"
-        },
-        "windowsOsState": {
-            "type": "string",
-            "allowedValues": [
-                "NonSysprepped",
-                "SysprepRequested",
-                "SysprepApplied"
-            ],
-            "defaultValue": "NonSysprepped"
+        "vmOsInfo": {
+            "type": "object"
         },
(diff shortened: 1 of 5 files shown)


VstsTasks/AzureDtlCreateCustomImage/new-azuredtl-customimage.json
```diff
@@ -8,34 +8,8 @@
         "labName": {
             "type": "string"
         },
```
ajay-arora: github makes it really hard to understand the diff ... because we dont see the entire files here. Could you describe what was the bug, so that we can understand the fix.

VstsTasks/AzureDtlCreateCustomImage/new-azuredtl-customimage.json
```diff
@@ -59,16 +33,7 @@
...
-                "vm": {
-                    "sourceVmId": "[parameters('sourceLabVmId')]",
-                    "linuxOsInfo":  {
-                        "linuxOsState": "[parameters('linuxOsState')]"
-                    },
-                    "windowsOsInfo": {
-                        "windowsOsState": "[parameters('windowsOsState')]"
-                    }
-                }
+                "vm": "[parameters('vmOsInfo')]"
```
ajay-arora: why did we remove the `osType` parameter that we were passing earlier.

leovms: We don't use it. Only when VHD is used.

VstsTasks/AzureDtlCreateCustomImage/new-azuredtl-customimage.json
leovms: Before we used the `osType` to distinguish in the RP and choose either the `linuxOsState` or `windowsOsState`. We have since moved to simply check if the OS state object is present and default to the Windows OS State check first. So, if both objects are passed, we take Windows even though the OS Type was set to Linux. Therefore, I have change the logic in the task to ensure we only pass the corresponding OS State based on the OS Type. We do the same thing from the Azure Portal logic. I match that logic.

PR #302, opened 2017-11-01
 added ServiceFabric environment template

markusheiliger:


ARMTemplates/ServiceFabric-LabCluster/metadata.json
```diff
@@ -0,0 +1,5 @@
+{
+  "itemDisplayName": "Service Fabric Lab Cluster",
+  "description": "This template allows you to deploy a secure 5 node, Single Node Type Service fabric Cluster running Ubuntu 16.04 or Windows Server 2016 on Standard_D2_v2 Size VMs with Windows Azure diagnostics turned on.",
+  "summary": "..."
```
leovms: Would it make sense to remove `summary` altogether or just make it an empty string, instead of an ellipsis?

PR #304, opened 2017-11-02
Fixing service init/start incompatibility with SLES 11

leovms:
Adding logic to fall back to "System V" to initialize services, instead of relying purely on "systemd", which is only installed on later flavors of Linux.


ericame:
LGTM

PR #305, opened 2017-11-02
Updating environment for Service Fabric Cluster.

EMaher:


EMaher:
Cleaning up the metadata file, adding script for creating needed certificate.  Also renamed ARMTemplates folder to Environments per to clarify what these ARM templates should be used for.

PR #306, opened 2017-11-04
Improving PS3 artifact resilience

leovms:
- Adding checks for failure scenarios to make artifact more resilient.
- Ensuring reboot happens adequately to allow applying subsequent artifacts in the same run.

### DIFF
--- a/Artifacts/windows-powershell3/install.cmd
+++ b/Artifacts/windows-powershell3/install.cmd
@@ -1,4 +1,4 @@
-@echo OFF
+@echo off
 
 rem This is required to ensure we don't get error 0x8024800C -2145091572 WU_E_DS_LOCKTIMEOUTEXPIRED The data store section could not be locked within the allotted time.
 echo Cleaning software distribution folder.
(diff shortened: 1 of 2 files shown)


Artifacts/windows-powershell3/install.cmd
```diff
@@ -1,14 +1,45 @@
 @echo OFF
 
-reg Query "HKLM\Hardware\Description\System\CentralProcessor\0" | find /i "x86" > NUL && set OSVERSION=32BIT || set OSVERSION=64BIT
+rem This is required to ensure we don't get error 0x8024800C -2145091572 WU_E_DS_LOCKTIMEOUTEXPIRED The data store section could not be locked within the allotted time.
+echo Cleaning software distribution folder.
+net stop "Windows Update" >nul 2>&1
+echo "Y" | del /f /s C:\Windows\SoftwareDistribution\ >nul 2>&1
+net start "Windows Update" >nul 2>&1
 
-if %OSVERSION%==32BIT (
```
workleodig: You could remove the "reg query" and simplify a bit with something like this:
if defined PROGRAMFILES(X86) (
REM do 64bit stuff
) else (
REM do 32bit stuff
)

Artifacts/windows-powershell3/install.cmd
```diff
@@ -1,14 +1,45 @@
...
+    goto EXIT
+)
+goto EXIT_ERROR
+
+:EXIT_ERROR
+echo An error occured during artifact installation. Review the virtual machine logs for more details. (%ERRORLEVEL%)
+exit %ERRORLEVEL%
+
+:RESTART
+rem Two restarts are required. This is the first one. The second one is defined in the postDeployActions of the artifactfile.json.
```
workleodig: I ran into a problem with these changes and one other artifact. 
I started with the Win2008 R2 marketplace image, and applied two artifacts when creating the VM - this artifact and a second one that has its commandToExecute set to "powershell.exe ..."
The second artifact failed saying "**'powershell.exe' is not recognized as an internal or external command, operable program or batch file**."
If I log into the VM after I get this error I don't have any trouble accessing powershell. Perhaps the CSE is getting confused by the new powershell version?

leovms: These changes are relying on some other back-end changes currently under development. With those chances in place, the logic is better handled. I will re-validate to confirm. Thanks for the feedback.

Artifacts/windows-powershell3/install.cmd
leovms: Yep. I'll ask the OP if there was a reason to run the query. Thanks.

leovms: No real reason. So, I made the change. It seemed to work fine in Windows Server 2008 R2 SP1.

PR #309, opened 2017-11-08
Add lab virtual machine owner re-esablishment script and readme

ryanvog:


Scripts/ResetVmPermissions/FixRoleAssignments.ps1
```diff
@@ -0,0 +1,94 @@
...
+    [Parameter(Mandatory=$false, HelpMessage="If you want to optionally target specific subscriptions instead of all of them, pass them in as a set")]
+    [Array] $SubscriptionIds,
+
+    [Parameter(Mandatory=$false, HelpMessage="If we should ALSO give VM Owners rights to the DevTest Lab, specify the Lab role - either DevTest Lab User, or Owner")]
+    [ValidateSet("DevTest Labs User", "Owner", "")]
+    [string] $LabRoleToAdd = ""
+)
+
+if ($SubscriptionIds -eq $null) {
+    $SubscriptionIds  = Get-AzureRmSubscription
```
leovms: Qualify the variable to ensure it is always constructed as an array, even when only 1 record is returned.
```powershell
[Array] $SubscriptionIds = Get-AzureRmSubscription
```

Scripts/ResetVmPermissions/FixRoleAssignments.ps1
```diff
@@ -0,0 +1,94 @@
...
+    # select the subscription
+    $sub = Select-AzureRmSubscription -SubscriptionId $subscription
+
+    if ($sub -eq $null) {
+        Write-Output "Unable to find any subscriptions.  Perhaps you need to run 'Add-AzureRmAccount' and login before running this script? Unable to proceed."
+        return
+    }
+
+    # Give me all labs in the subscription
+    $devTestLabs = Find-AzureRmResource -ResourceType 'Microsoft.DevTestLab/labs'
```
leovms: Ditto about `Array`. It is definitely precautionary.

Scripts/ResetVmPermissions/FixRoleAssignments.ps1
```diff
@@ -0,0 +1,94 @@
...
+            $ownerName= $virtualMachine.Properties.ownerUserPrincipalName
+
+            # the owner can be blank for really old VMs, let's copy created by if machine isn't claimable
+            if (($virtualMachine.Properties.allowClaim -eq $false) -and ($ownerId -eq $null)) {
+                $ownerId = $virtualMachine.Properties.createdByUserId
+                $ownerName = $virtualMachine.Properties.createdByUser
+            }
+
+            # if we have the owner Id, let's check the permissions
+            if (($ownerId -ne $null) -and ($ownerId.Trim().Length -gt 0) ) {
```
leovms: You can also use the following:
```powershell
if (-not [String]::IsNullOrWhitespace($ownerId))
```

Scripts/ResetVmPermissions/FixRoleAssignments.ps1
```diff
@@ -0,0 +1,94 @@
...
+                
+                $parentResource = ("labs/" + $devTestLab.Name)
+
+                # If the VM role assignment doesn't exist, let's add it back
+                $existing = Get-AzureRmRoleAssignment -ObjectId $ownerId `
+                                                      -RoleDefinitionName "Owner" `
+                                                      -ResourceGroupName $virtualMachine.ResourceGroupName `
+                                                      -ResourceName $virtualMachine.Name `
```
leovms: Even though you're specifying `ParentResource`, you may want to double check, if not already done, the resource name, as the format, if I recall, is `labs/virtualmachines`. Therefore, the name may be `labName/virtualMachineName`.

Scripts/ResetVmPermissions/FixRoleAssignments.ps1
```diff
@@ -0,0 +1,94 @@
...
+                    $temp = New-AzureRmRoleAssignment -ObjectId $ownerId `
+                                                      -RoleDefinitionName "Owner" `
+                                                      -ResourceGroupName $virtualMachine.ResourceGroupName `
+                                                      -ResourceName $virtualMachine.Name `
+                                                      -ResourceType $virtualMachine.ResourceType `
+                                                      -ParentResource $parentResource
+                }
+
+                # if the Lab role assignment doesn't exist, and we're supposed to add it, let's do it
+                if ($LabRoleToAdd -ne "") {
```
leovms: Could you not combine the above "if ($existing -eq $null)" and set `$LabRoleToAdd` to default to `Owner`? That is, remove the blank value from the set and just default to `Owner`. Just wondering if the two parts of the logic, which seem very similar, could not be combined.

ryanvog:
Thanks @leovms , I will push this in now and update based on your feedback as a follow up.

PR #316, opened 2017-11-22
Add auto-shutdown verification script

ryanvog:


Scripts/GetAutoShutdownPolicyStatus/GetAutoShutdownPolicyStatus.ps1
```diff
@@ -0,0 +1,98 @@
...
+
+                # Get the tag and use it to determine AS status
+                $tags = $virtualMachine.Tags;
+
+                if ($tags -and $tags["AlwaysOn"] -eq "true") {
+                    Write-Warning "   >>>> Virtual machine $($virtualMachine.Name) auto-shutdown status is disabled"
+                    $labHealthy = $false
+                }
+
+                $vmSchedule = $null
```
joerage: Skip the rest of this VM logic if the 'AlwaysOn' is set? Otherwise, you may end up with outputting the same VM twice, no?

Scripts/GetAutoShutdownPolicyStatus/GetAutoShutdownPolicyStatus.ps1
```diff
@@ -0,0 +1,98 @@
...
+            $schedDisabled = $true
+            $labHealthy = $false
+        }
+
+        if (-not $schedDisabled) {
+            # Get all the VMs in the lab for an override
+            $virtualMachines = Get-AzureRmResource -ResourceId "$($devTestLab.ResourceId)/virtualmachines" -ApiVersion 2016-05-15 
+
+            foreach ($virtualMachine in $virtualMachines) {
+                Write-Verbose "   >>> Processing virtual machine $($virtualMachine.Name)"
```
joerage: If someone is looking at a quick summary to know which VMs are not opted in, this line will polute the output. If there is 50 VMs in the lab for example, user will have to look closely to find what VM is disabled.

joerage: Oh, nevermind, I see this is only with the -verbose switch turned on.

Scripts/GetAutoShutdownPolicyStatus/GetAutoShutdownPolicyStatus.ps1
```diff
@@ -0,0 +1,98 @@
...
+        }
+        catch {
+            Write-Warning "Lab $($devTestLab.Name) auto-shutdown policy is disabled or no longer present"
+            $labHealthy = $false
+            continue
+        }
+
+        $schedDisabled = $false
+
+        if (($schedule -eq $null) -or ($schedule.Properties.status -ne 'Enabled')) {
```
leovms: You can collapse this check, as PS never throws when `$null` or undefined.
```powershell
if ($schedule.Properties.status -ne 'Enabled')
```

Scripts/GetAutoShutdownPolicyStatus/GetAutoShutdownPolicyStatus.ps1
```diff
@@ -0,0 +1,98 @@
...
+    [String] $LabName
+
+)
+
+$console = (Get-Host).PrivateData
+$console.VerboseForegroundColor = "Gray"
+$console.WarningForegroundColor = "Yellow"
+
+if ($SubscriptionIds -eq $null) {
+    $SubscriptionIds = Get-AzureRmSubscription
```
leovms: Change to the following to isolate the subscription IDs.
```powershell
[Array] $SubscriptionIds = Get-AzureRmSubscription | Select-Object -Property Id
```

Scripts/GetAutoShutdownPolicyStatus/README.md
```diff
@@ -0,0 +1,20 @@
+# Introduction
+
+When you manage multiple labs, it is sometimes important to make sure that all your labs have the Auto-Shutdown policy enabled.  This can be a tedious exercise if you have many labs to examine.  To automate this process, we have introduced a PowerShell script that will look at all your labs and verify they are opted-in to the Auto-Shutdown lab policy. 
+
+The syntax of the script is as follows:
+```
```
leovms: Add `powershell` after ` ``` `, as in ` ```powershell`, to provide syntax highlighting for the examples.

For example,
```
` ` `powershell
GetAutoShutdownPolicyStatus.ps1 [-SubscriptionIds] [-LabName] [-Verbose]
` ` `
```
turns into
```powershell
GetAutoShutdownPolicyStatus.ps1 [-SubscriptionIds] [-LabName] [-Verbose]
```

Scripts/GetAutoShutdownPolicyStatus/GetAutoShutdownPolicyStatus.ps1
```diff
@@ -0,0 +1,98 @@
...
+
+        if (($schedule -eq $null) -or ($schedule.Properties.status -ne 'Enabled')) {
+            Write-Warning "Lab $($devTestLab.Name) auto-shutdown policy is disabled"
+            $schedDisabled = $true
+            $labHealthy = $false
+        }
+
+        if (-not $schedDisabled) {
+            # Get all the VMs in the lab for an override
+            $virtualMachines = Get-AzureRmResource -ResourceId "$($devTestLab.ResourceId)/virtualmachines" -ApiVersion 2016-05-15
```
leovms: Strong type the variable to ensure it is an array even when only one record is returned, as in
```powershell
[Array] $virtualMachines = ...
```

Scripts/GetAutoShutdownPolicyStatus/GetAutoShutdownPolicyStatus.ps1
```diff
@@ -0,0 +1,98 @@
...
+                $vmSchedule = $null
+
+                try {
+                    $vmSchedule = Get-AzureRmResource -ResourceId "$($devTestLab.ResourceId)/virtualmachines/$($virtualMachine.Name)/schedules/LabVmsShutdown" -ApiVersion 2016-05-15
+                }
+                catch {
+                        
+                }
+    
+                if (($vmSchedule -ne $null) -and ($vmSchedule.Properties.status -ne 'Enabled')) {
```
leovms: Collapse condition.
```powershell
if ($vmSchedule.Properties.status -ne 'Enabled')
```

ryanvog: I need the null check here because it was a precondition that I check afterwards.  Removing the null means I will have to change the way I work with the 'no schedule found' scenario.   Leaving this one as is for now.

Issue #318, opened 2017-11-28
VSTS Build Agent Artifact does not install in Windows Server 1709


rfennell:
I have tried in DevLabs to create a formula based on a 'Windows Server 1709' VM on an A0 VM, to this I have added the 'VSTS Build Agent' artifact. 

The formula deploys the VM OK, but the artifacts fails.

When I RDP into the VM I can see the new root folder (with the same name as the VM) is created by the artifact, (off the build agent's root). However the expansion of the downloaded agent seems to have failed.

The log is as follows

> Validating parameters
> Preparing agent installation location
> Checking for previously configured agent
> Downloading agent package
> Extracting agent package contents
> ERROR: Error HRESULT E_FAIL has been returned from a call to a COM component.

I get the same error if try to apply the artifact the a previously deployed VM. So it is not related to initial VM creation.

I checked I can deploy and configure the agent manually as I detailed in [this post](https://blogs.blackmarble.co.uk/rfennell/2017/11/23/creating-a-vsts-build-agent-on-an-azure-devlabs-windows-server-vm-with-no-gui/), so I know the issue is not a lack of VM resources.

I am at a bit of a loss why there is a difference between a manual deploy and an artifact based one.

rfennell:
Closing a PR has been accepted

PR #319, opened 2017-11-28
VSTS Build Agent artifact Swapped to System.IO.Compression.FileSystem to extract file

rfennell:
This is a fix for issue #318 - that the VSTS build agent artifact fails to deploy on a non GUI Windows Server VM.

Swapped to using **System.IO.Compression.FileSystem** as opposed to **New-Object -ComObject shell.application** call.

I have tested this on using my DevTest Labs instance using my fork as a private repo and it deploys without error

### DIFF
--- a/Artifacts/windows-vsts-build-agent/vsts-agent-install.ps1
+++ b/Artifacts/windows-vsts-build-agent/vsts-agent-install.ps1
@@ -220,9 +220,10 @@ function Extract-AgentPackage
         [string] $PackagePath,
         [string] $Destination
     )
+  
+    Add-Type -AssemblyName System.IO.Compression.FileSystem 
+    [System.IO.Compression.ZipFile]::ExtractToDirectory("$PackagePath", "$Destination")
     
-    $destShellFolder = (New-Object -ComObject shell.application).namespace("$Destination")
-    $destShellFolder.CopyHere((New-Object -ComObject shell.application).namespace($PackagePath).Items(), 16)
 }
 
 function Install-Agent


Artifacts/windows-vsts-build-agent/vsts-agent-install.ps1
```diff
@@ -220,9 +220,10 @@ function Extract-AgentPackage
         [string] $PackagePath,
         [string] $Destination
     )
+
```
leovms: Can you clarify if your changes are backward compatible? I assume that the original author may have used that approach to potentially support earlier versions of Windows or PS or the like.

Do we know if we may need to add a condition around using one approach over another? I suspect that if we validate against Windows Server 2008 R2 SP1, or similar, we should be OK. Can you please confirm?

rfennell: I will spin up a Windows Server 2008 R2 SP1 hosted build agent and give it a try and get back to you

rfennell: OK, I have done some testing.

In DevTest Labs I deployed a new **Windows Server 2008 R2 SP1** from the standard base image and the **publicly available** version of  the **deploy a build agent artifact** from the **Official Microsoft Repo**. The VM deployed but the artifact failed, I got the error:

> The term 'Handle-LastError' is not recognized as the name of a cmdlet, function
> , script file, or operable program. Check the spelling of the name, or if a pat
> h was included, verify that the path is correct and try again.
> At C:\Packages\Plugins\Microsoft.Compute.CustomScriptExtension\1.9\Downloads\2\
> 2008R2SP1.131565418593395186\windows-vsts-bu1072864756\scripts\vsts-agent-insta
> ll.ps1:271 char:21
> +     Handle-LastError <<<< 
>     + CategoryInfo          : ObjectNotFound: (Handle-LastError:String) [], Co 
>    mmandNotFoundException
>     + FullyQualifiedErrorId : CommandNotFoundException
>  

Incidentally I got the same error using the new version of the artifact from my fork'ed repo, but I would expect that as this looks like a basic PowerShell version issue i.e: just the concern you raised, but at a more fundamental level - the whole script fails not my replacement lines.

So after this initial test I put the problems down to the default PowerShell Version 2 in Windows Server 2008 R2 SP1.

To try to fix this, I deployed the PowerShell 3 artifact to the previously deployed Windows Server 2008 R2 SP1 VM. This deployed with no errors, and caused no reboot.

I tried to deploy the updated build agent artifact from **my fork'd repo** - and it worked, but after a bit of thought I realised I had triggered a rebooted the VM after PowerShell was installed. 

So I then did an end to end check, deploying a new **Windows Server 2008 R2 SP1 VM** with the **PowerShell 3 with Windows Management** artifact and **(my fork of) the Build Agent artifact**. This new deployment deployed PowerShell (with no reboot), but failed deploying the build agent artifact, this time with a new the error

> 'powershell.exe' is not recognized as an internal or external command,
> operable program or batch file.

I next tried to redeloy the build artifact (still without a reboot to see if it was timing issue) but got the same error.

I then did  a VM reboot and retried the same build agent artifact deploy and it worked, deploying and registering the agent on my VSTS instance.

**So not sure what you want to do here, my code change does require at least PowerShell 3, but so does your currently shipped version of the artifact**

leovms: Thank you for the thorough testing. The issue with PS3 is something that I've already fixed in a separate fork, which relies on some backend changes. I will update that separately and when the backend changes happen.

All good from my POV. Appreciate you going through it.

PR #323, opened 2017-12-04
Updated logic used to validate artifact status

leovms:
Updated logic used to validate artifact status used to determine when to fail the task.

### DIFF
--- a/VstsTasks/AzureDtlCreateVM/task-funcs.ps1
+++ b/VstsTasks/AzureDtlCreateVM/task-funcs.ps1
@@ -237,8 +237,8 @@ function Validate-ArtifactStatus
     if ($checkForFailedArtifacts)
     {
         $templateFile = Get-TemplateFile -TemplateName $TemplateName
-        # Read the contents of the ARM template and remove any comments in the form /* ... */
-        # since these cause the call to ConvertFrom-Json later on fail.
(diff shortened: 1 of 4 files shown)


VstsTasks/AzureDtlCreateVM/task-funcs.ps1
```diff
@@ -205,18 +229,37 @@ function Validate-ArtifactStatus
...
-    if ($fail)
+    $checkForFailedArtifacts = ConvertTo-Bool -Value $Fail
+    if ($checkForFailedArtifacts)
     {
-        $artifactDeploymentStatus = (Get-AzureRmResource -ResourceId $ResourceId -ODataQuery '$expand=Properties($expand=Artifacts)').Properties.artifactDeploymentStatus
-
-        [array]$failedArtifacts = $artifactDeploymentStatus | ? { $_.deploymentStatus -eq 'Failed' }
-        if ($failedArtifacts.Count -gt 0)
+        $templateFile = Get-TemplateFile -TemplateName $TemplateName
+        $armTemplateJson = [IO.File]::ReadAllText($templateFile) -replace '/\*([^\*/])*\*/',''
```
drwill-ms: Add a code comment as to what this is doing, which I assume is removing any code comments in the style of: /* ... */.

Also, perhaps we should not keep this in long term, eh?

PR #325, opened 2017-12-06
Changing artifact to use PowerShell's -Command

leovms:
Changing artifact to use PowerShell's -Command to bubble up errors correctly.


joerage:
Good.

PR #332, opened 2018-01-11
Improving retry logic

leovms:
Adding flags and corresponding logic to control deletion of a failed deployment before retrying and appending the retry instance number to the VM name.

### DIFF
--- a/VstsTasks/AzureDtlCreateVM/task.ps1
+++ b/VstsTasks/AzureDtlCreateVM/task.ps1
@@ -29,7 +29,9 @@ param(
     [string] $OutputResourceId,
     [string] $FailOnArtifactError,
     [string] $RetryOnFailure,
-    [string] $RetryCount
+    [string] $RetryCount,
+    [string] $DeleteFailedDeploymentBeforeRetry,
+    [string] $AppendRetryNumberToVMName
 )
 
 ###################################################################################################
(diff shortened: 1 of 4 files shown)


VstsTasks/AzureDtlCreateVM/task.ps1
```diff
@@ -117,14 +128,10 @@ try
...
-                {
-                    Remove-AzureRmResource -ResourceId $resourceId -Force | Out-Null
-                }
-                else
+                Remove-FailedResourcesBeforeRetry -DeploymentName $deploymentName -ResourceGroupName $resourceGroupName -DeleteDeployment $DeleteFailedDeploymentBeforeRetry
+                if ($AppendRetryNumberToVMName)
                 {
-                    Write-Host "Resource identifier is not available, will not attempt to remove corresponding resouce before retrying."
-                    Write-Host "Dumping raw deployment result:`n$result"
+                    $templateParameterObject.newVMName = "$baseVmName-$i"
```
ericame: Do we need to worry about VM name length at this point?

leovms: We do have logic to check for that already. In the flag description we do mention that enabling this flag may increase and therefore fail VM name limits.

PR #333, opened 2018-01-12
Remove yum update line

grwann:
Running yum update then [updates ](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/update-infrastructure-redhat) the version of Red Hat and breaks deprovision.  Removed this line as to not cause side effects when running deprovision.

Have tested on linux flavors with yum that existing functionality is not broken.

### DIFF
--- a/Artifacts/linux-deprovision/depro.sh
+++ b/Artifacts/linux-deprovision/depro.sh
@@ -11,7 +11,6 @@ main () {
             apt-get -y update > /dev/null
             apt-get -y install at > /dev/null
         elif [ -n "$isYum" ] ; then
-            yum -y update > /dev/null
             yum install -y at > /dev/null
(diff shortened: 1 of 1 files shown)


Artifacts/linux-deprovision/depro.sh
```diff
@@ -11,7 +11,6 @@ main () {
             apt-get -y update > /dev/null
             apt-get -y install at > /dev/null
         elif [ -n "$isYum" ] ; then
-            yum -y update > /dev/null
```
leovms: What's the outcome of not running this line?

grwann: It should all work as expected, I'm testing this script on every family of linux we have to make sure it doesn't break anything.

PR #336, opened 2018-01-18
Add install certificate artifact

ross-bohr:
Added a new artifact which is able to install a certificate to the VM. It accepts the following parameters:

1. Certificate Name [string] - used to distinguish between multiple instances of that artifact, in case you need to install more than one certificate.
2. Certificate in Base64 [string] - a base64 encoded form of the certificate's PFX file.
3. Certificate password [secure string] - the password for the PFX.

The artifact installs the certificate and reports any progress to the Azure blade console.


Artifacts/windows-install-certificate/Artifactfile.json
```diff
@@ -0,0 +1,35 @@
...
+  "targetOsType": "Windows",
+  "parameters": {
+    "certificateName": {
+      "type": "string",
+      "displayName": "Certificate Name",
+      "description": "The name of the certificate which must be installed. This property is used to distinguish multiple instances of this artifact.",
+      "defaultValue": "NewCertificate"
+    },
+    "base64cert": {
+      "type": "string",
```
leovms: You could consider changing the type to `fileContents` and it would allow you to upload a file instead. Just another possibility. Take a look at the artifact [`Test Parameter Types`](https://github.com/Azure/azure-devtestlab/tree/master/Artifacts/windows-test-paramtypes).

Artifacts/windows-install-certificate/certificate-install.ps1
```diff
@@ -0,0 +1,109 @@
...
+    Write-Host "Installing certificate $certificateName"
+
+    If(-NOT ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator"))
+    {
+        #build up the deploy arguments
+        $arguments = "-file `"{0}`"" -f $script:MyInvocation.MyCommand.Path
+    
+        # Start the new process
+        Start-Process powershell.exe -Verb runas -ArgumentList $arguments
+        exit
```
leovms: Do you really need to call `exit` here? The call to `Write-Host 'Done'` at the end will not execute.

Also, should we consider adding `-Wait` here? Otherwise, the process will alway be launched but you won't get any feedback if it fails. So, the artifact will be set to `Succeeded` potentially prematurely.

Artifacts/windows-install-certificate/certificate-install.ps1
```diff
@@ -0,0 +1,109 @@
...
+    }
+    else
+    {
+        $securePassword = ConvertTo-SecureString -String $certificatePassword -AsPlainText -Force
+		$certificatePassword = "deleted"
```
leovms: Remove indent.

Artifacts/windows-install-certificate/certificate-install.ps1
```diff
@@ -0,0 +1,109 @@
...
+        $tempFilePath = [System.IO.Path]::GetTempFileName()
+        Write-Host "Temp file path '$tempFilePath'" 
+
+        [System.IO.File]::WriteAllBytes($tempFilePath, [System.Convert]::FromBase64String($base64cert))
+        Write-Host "Certificate saved"
+        		
+        Get-ChildItem -Path $tempFilePath | Import-PfxCertificate -CertStoreLocation Cert:\LocalMachine\My -Exportable -Password $securePassword
+        Write-Host "Certificate $certificateName added to the LocalMachine\My store succesfully"
+
+        [System.IO.File]::Delete($tempFilePath)
```
leovms: Consider using `Remove-Item -Path "$tempFilePath" -Force`, as it will not throw.

Artifacts/windows-install-certificate/certificate-install.ps1
ross-bohr: This part of the script is not specific to this artifact. It can be copied from any of the MS provided artifacts and I don't want to change the model.

Artifacts/windows-install-certificate/Artifactfile.json
ross-bohr: It is convenient to store the pfx file in base64 string together with the artifact instance in a DevTest Lab formula instance. This is the same approach MS follows for Service Fabric cluster connectors in VSTS, thus a standard one. I'd not change it.

Artifacts/windows-install-certificate/certificate-install.ps1
ross-bohr: Implemented in the latest commit.

Artifacts/windows-install-certificate/certificate-install.ps1
ross-bohr: Implemented in the latest commit.

Artifacts/windows-install-certificate/Artifactfile.json
ross-bohr: azure_devtestlab_formula
A sample formula in Azure DevTest Lab as seen in Azure Resource Explorer.

ross-bohr: It would be nice to use "fileContents", however it seems it has the following limitation:
"Validation error for mycert.pfx. Details: Selected file 'mycert.pfx' must be between 1 and 2048 bytes."

Artifacts/windows-install-certificate/certificate-install.ps1
ross-bohr: Implemented in the latest commit.

Artifacts/windows-install-certificate/Artifactfile.json
ross-bohr: @leovms  , is there a possibility to override the default size limit (1-2048 bytes) for "fileContents" ? If not, then it is not practical to use this type.

ross-bohr:
@leovms 

Could you review the updated changes and provide an answer to the "fileContents" type question above?

ross-bohr:
@leovms 

Could you please answer the above question?

ross-bohr:
@jaymarf @mithunshanbhag @mmydland 

Could you kindly have a look at the above PR and questions?

Artifacts/windows-install-certificate/Artifactfile.json
leovms: No, it seems we're limited to that size.

leovms:
@ross-bohr, apologies for such a long delay. I'm reviewing all submissions.

PR #337, opened 2018-01-18
Add message in description for CoreOS

grwann:
CoreOS deprovision does not work, and it is not a very popular flavor in DTL so we will postpone the investment to support it.  We however will add a warning in the artifact description so potential customers will know.

### DIFF
--- a/Artifacts/linux-deprovision/Artifactfile.json
+++ b/Artifacts/linux-deprovision/Artifactfile.json
@@ -1,7 +1,7 @@
 {
     "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2016-11-28/dtlArtifacts.json",
     "title": "de-provision",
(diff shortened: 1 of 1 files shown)


Artifacts/linux-deprovision/Artifactfile.json
```diff
@@ -1,7 +1,7 @@
 {
     "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2016-11-28/dtlArtifacts.json",
     "title": "de-provision",
-    "description": "Uses the Microsoft Azure Linux Agent (waagent) to de-provision a Linux VM for saving as an image.",
+    "description": "Uses the Microsoft Azure Linux Agent (waagent) to de-provision a Linux VM for saving as an image.  This artifact does not work on the Core OS flavor of Linux..",
```
leovms: Remove extra period at the end of the description.

PR #341, opened 2018-01-25
Fix domain join null argument exception

samedder:
The command `Add-Computer` expects non-null arguments when invoked. When calling windows domain join, this can result in a null argument exception when the OUPath is not specified.

Related to previous change #340.

### DIFF
--- a/Artifacts/windows-domain-join-new/artifact.ps1
+++ b/Artifacts/windows-domain-join-new/artifact.ps1
@@ -43,8 +43,11 @@ function Add-VmToDomain ()
     else
     {
         $credential = New-Object System.Management.Automation.PSCredential($JoinUser, $JoinPassword)
-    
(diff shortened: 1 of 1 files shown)


Artifacts/windows-domain-join-new/artifact.ps1
```diff
@@ -68,10 +71,6 @@ else
 {
     $securePass = ConvertTo-SecureString $DomainAdminPassword -AsPlainText -Force
     Write-Output "Attempting to join the domain..."
-    # Treat empty string as null to convert to optional parameter
-    if(-not $OUPath) {
-        $OUPath = $null
-    }
     Add-VmToDomain -VmName $env:COMPUTERNAME -DomainName $DomainToJoin -JoinUser $DomainAdminUsername -JoinPassword $securePass -OU $OUPath
```
leovms: `$OUPath` doesn't exist anymore, from what I can see. What about the `-OU` parameter in this call?

PR #346, opened 2018-02-02
Support for multiple build artifacts and improved resiliency

leovms:
- Added support for downloading multiple build artifacts.
- Throwing when missing data to stop processing and bubble errors.

### DIFF
--- a/Artifacts/windows-vsts-download-and-run-script/DownloadVstsDropAndExecuteScript.ps1
+++ b/Artifacts/windows-vsts-download-and-run-script/DownloadVstsDropAndExecuteScript.ps1
@@ -1,6 +1,5 @@
 [CmdletBinding()]
 param(
-    [string] $username,
     [string] $accessToken,
     [string] $buildDefinitionName,
     [string] $vstsProjectUri,
@@ -9,7 +8,6 @@ param(
 )
 
 ###################################################################################################
-
 #
 # PowerShell configurations
 #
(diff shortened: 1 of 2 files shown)


Artifacts/windows-vsts-download-and-run-script/DownloadVstsDropAndExecuteScript.ps1
```diff
@@ -49,16 +44,52 @@ function Handle-LastError
...
+        Remove-Item -Path $Destination -Force -Recurse | Out-Null
+    }
+
+    Write-Host "Getting build artifacts information from $ArtifactsUri"
+    [Array] $artifacts = (Invoke-RestMethod -Uri $ArtifactsUri -Headers $Headers -Method Get | ConvertTo-Json -Depth 3 | ConvertFrom-Json).value
+
+    # Process all artifacts found.
+    foreach ($artifact in $artifacts)
+    {
+        $artifactName = "$($artifact.name)"
```
drwill-ms: Couldn't "$($artifact.name)" be simplified to $artifact.name?

EMaher: Unfortunately, "$($artifact.name)" cannot be simplified to $artifact.name.  PowerShell supports expanding variables or expressions, but expressions (even ones that just access a property on a variable) must be enclosed in parenthesis.  The PowerShell team wrote an interesting blog explaining [variable expansion and Here-Strings](https://blogs.msdn.microsoft.com/powershell/2006/07/15/variable-expansion-in-strings-and-here-strings/).

leovms: Like Elizabeth said. Additionally, because the artifact name can have spaces, we want to make sure they are maintained. Although, just putting `$artifact.name` should also work.

PR #349, opened 2018-02-13
Fixed regression in autologon due to azure vms not creating the user hive until first rdp or login

ShreyasRmsft:


ShreyasRmsft:
@leovms Please review

ShreyasRmsft:
@leovms please review

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
...
+        if ($windowsLogonAccount.Split("\").Count -eq 2)
+        {
+            $domain = $windowsLogonAccount.Split("\")[0]
+        }
+        else
+        {
+            $domain = Hostname
+        }
+        $computerName = "localhost"
+        $password = ConvertTo-SecureString $windowsLogonPassword -AsPlainText -Force
```
leovms: Where do you validate that the parameter `$windowsLogonPassword` is not `$null` or empty? If you call `ConvertTo-SecureString` with a `$null` value it will throw. I see that we're checking for such condition later on in this flow, but you should consider rearranging that logic.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
 
     if ($runAsAutoLogon)
     {
+        # Create a PS session for the user to trigger the creation of the registry entries required for autologon
+        if ($windowsLogonAccount.Split("\").Count -eq 2)
```
leovms: This change means that we can no longer use an email address to authenticate. Am I to assume that's expected?

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
...
+        {
+            $domain = $windowsLogonAccount.Split("\")[0]
+        }
+        else
+        {
+            $domain = Hostname
+        }
+        $computerName = "localhost"
```
leovms: Will `localhost` always work? Do you need to consider something like `"$($Env:ComputerName)"`?

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
...
+        $password = ConvertTo-SecureString $windowsLogonPassword -AsPlainText -Force
+        $credentials = New-Object System.Management.Automation.PSCredential("$domain\\$windowsLogonAccount", $password)
+        Enter-PSSession -ComputerName $computerName -Credential $credentials
+        Exit-PSSession
+
+        $ErrorActionPreference = "stop"
```
leovms: This should be moved to the start of the script and should apply to everything, not just in this context.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
...
+
+        try
+        {
+            # Check if the HKU drive already exists
+            Get-PSDrive -PSProvider Registry -Name HKU | Out-Null
+            $canCheckRegistry = $true
+        }
+        catch [System.Management.Automation.DriveNotFoundException]
+        {
+            try
```
leovms: Fix indentation in this `try`/`catch`.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
...
+        }
+        $computerName = "localhost"
+        $password = ConvertTo-SecureString $windowsLogonPassword -AsPlainText -Force
+        $credentials = New-Object System.Management.Automation.PSCredential("$domain\\$windowsLogonAccount", $password)
+        Enter-PSSession -ComputerName $computerName -Credential $credentials
+        Exit-PSSession
+
+        $ErrorActionPreference = "stop"
+
+        # 120 seconds timeout
```
leovms: Move `$timeout` definition closer to the `while` loop.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
...
+            {
+            # Ignore the failure to create the drive and go ahead with trying to set the agent up
+            Write-Warning "Moving ahead with agent setup as the script failed to create HKU drive necessary for checking if the registry entry for the user's SId exists.\n$_"
+            }
+        }
+
+        # Check if the registry key required for enabling autologon is present on the machine, if not wait for 120 seconds in case the user profile is still getting created
+        while ($timeout -ge 0 -and $canCheckRegistry)
+        {
+            $objUser = New-Object System.Security.Principal.NTAccount($windowsLogonAccount)
```
leovms: Should you put all this logic in a `try`/`catch`? Because you have `$ErrorActionPreference = "stop"`, any issues that happen here will force the script to stop. It may not be what you want.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
```diff
@@ -144,6 +144,76 @@ function Configure-DeploymentGroupAgent
...
     if ($runAsAutoLogon)
     {
+        # Create a PS session for the user to trigger the creation of the registry entries required for autologon
+        if ($windowsLogonAccount.Split("\").Count -eq 2)
+        {
+            $domain = $windowsLogonAccount.Split("\")[0]
+        }
+        else
+        {
+            $domain = Hostname
```
leovms: Where does `Hostname` get defined? Are you expecting to use the literal string or is this supposed to be a variable you defined somewhere? Not sure I get its origin.

Artifacts/windows-vsts-build-agent/vsts-agent-install.ps1
```diff
@@ -247,6 +247,78 @@ function Install-Agent
 
         if ($Config.RunAsAutoLogon)
         {
+            # Create a PS session for the user to trigger the creation of the registry entries required for autologon
```
leovms: A good part of this is the same logic as already reviewed earlier. Please consider those comments as applicable here as well.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: Added a check for this. The later bit of code that checks and adds the password to the command string is common to both flows (autologon or service) hence leaving it as it is.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: You either provide a local account or a domain joined account that exists on the vm. There is an else block that handles non domain accounts. The behavior has not changed and both cases are handled. When you say email address you mean a domain joined account right?

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: It has to be localhost. It doesn't work with $($Env:ComputerName) as this login has to bypass the NIC and go via loopback since azure VMs don't support pssessions via http. (only https supported)

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: done

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: Actually no. There is a crucial step New-Item -Path "HKU:\\$securityId\\SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Run" -Force, If this fails then the agent will get registered to tfs but won't start up. I would want the script to stop if this check or this creation fails. 

The only case in which I go ahead is if in case i somehow do not have permissions to check for this key then I do not have sufficient info to decide whether to stop or go ahead.

case in point being the previous catch block.

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: Its an alias for hostname.exe, changed it to the use the env var $Env:ComputerName

Artifacts/windows-vsts-build-agent/vsts-agent-install.ps1
ShreyasRmsft: done

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: did not want to change the behavior of code outside, needed it for my try catch blocks. But moved it to the beginning after checking the rest of the code to make sure it won't cause any behavior changes elsewhere

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: done

ShreyasRmsft:
@leovms replied to all comments and made necessary changes

Artifacts/windows-deploymentgroup/InstallAndConfigureDeploymentGroupAgent.ps1
ShreyasRmsft: Moved it back here as it had a few undesired effects interms of proper error messages being displayed. The initial part of the script requires that this not be set.  After that it's fine to have this. Didn't want to touch the existing code as the scenarios are too vast for me to manually test all possible success and failure scenarios

PR #389, opened 2018-06-28
Clean up .gitignore and remove tracked, ignored files

ChadNedzlek:
The .gitignore file seems to have some redundant and incorrect entries,
and only ignored Debug, not release directories.

Also, there are some files checked in that should be ignored, which
contain temporarily Visual Studio files that contain computer
information about the last person to commit them.


ChadNedzlek:
@ShreyasRmsft  @mmydland I don't really know what this sample does, but these files are preventing me from syncing because they have relatively long paths, and don't appear to be useful files. Also, they contain information that really shouldn't be checked in.

ShreyasRmsft:
@ChadNedzlek the service exe is uploaded to a public blob. It was put here in case we ever needed to go back and update.

Yeah those files are not required. Thanks.

PR #390, opened 2018-07-09
Adding in new script to update RDP extended properties

petehauge:
**This PR includes:**

- Readme.md :  Description of the script files
- Set-DTL_RemoteDesktopSettings.ps1 :  Version of the script that works with Powershell 5.
- Set-DTL_RemoteDesktopSettings_ps6.ps1:  Version of the script that works with Powershell 6.


Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
```diff
@@ -0,0 +1,90 @@
+﻿<#
+
+.SYNOPSIS
+This script leverages Azure Powershell to update the Remote Desktop settings in an Azure DevTest Lab.
+
+.DESCRIPTION
+The script updates the DevTest Lab's extended properties to include the experience level and/or the gateway settings.  When a user clicks "connect" on a Virtual Machine, the RDP file that's generated includes these settings.  In this way, IT admins or DevTest Lab Owners can configure the appropriate remote desktop settings for the environment and the uesrs can just leverage it seamlessly.
```
leovms: Not a very big issue, but we may want to consider breaking this up so that it wraps to make it easier to read.

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
```diff
@@ -0,0 +1,90 @@
...
+
+    [Parameter(HelpMessage="The server name of the remote desktop host.  This cane be a fully qualitified name or an IP address of the gateway")]
+    [string] $RemoteDesktopGateway,
+
+    [Parameter(HelpMessage="The remote desktop experience level")]
+    [ValidateSet(1, 2, 3, 4, 5, 6, 7)]
+    [int] $ExperienceLevel
+)
+
+function Update-Member ($labProperties, $fieldName, $fieldValue) {
```
leovms: Should we validate these parameters?

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
```diff
@@ -0,0 +1,90 @@
...
+            $labProperties.extendedProperties.$fieldName = $fieldValue
+        }
+        else {
+            # add a new property since it doesn't already exist
+            Add-Member -InputObject $labProperties.extendedProperties -MemberType NoteProperty -Name $fieldName -Value $fieldValue
+        }
+    }
+    else {
+        # Clear out the field since the parameter is null, remove the property if it exists
+        # NOTE:  There isn't a "Remove-Member" so we have to do some converstions to remove the field
```
leovms: converstions?

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
```diff
@@ -0,0 +1,90 @@
...
+            $lab.Properties.extendedProperties.PSObject.Properties.Remove($fieldName)
+        }
+    }
+}
+
+# Find the DevTest Lab
+$lab = Get-AzureRmResource -ResourceId (Find-AzureRmResource -ResourceNameEquals $DevTestLabName -ResourceType "Microsoft.DevTestLab/labs").ResourceId -ApiVersion "2017-04-26-preview"
+
+if ($lab -eq $null) {
+    Write-Error "Unable to find the Lab named $DevTestLabName in Subscription $SubscriptionId"
```
leovms: Use `throw` here to force the script to terminate. Unless you set `$ErrorActionPreference = 'Stop'`, calling `Write-Error` will not terminate the script.

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings_ps6.ps1
```diff
@@ -0,0 +1,90 @@
+﻿<#
```
leovms: Same comments apply to this script file as well. Consider finding a way to combine these 2 scripts into a single one. It makes it a lot easier to maintain.

If you need to maintain conditional logic for PS5 vs PS6, consider checking `$PSVersionTable.PSVersion.Major`.

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
petehauge: Done!

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
leovms: As clarified offline. No need to validate these parameters since their values drive the logic.

petehauge: I added a little validation in the function at least, thanks!

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
petehauge: Good catch, this was an old comment - i found a better way to do it but forgot to remove the comment.  Deleted now...

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings.ps1
petehauge: It should work fine because the 'write-error' is inside the if statement and after this line it branches to the end.  I switched it around so it's easier to read though...

Scripts/UpdateRemoteDesktopSettings/Set-DTL_RemoteDesktopSettings_ps6.ps1
petehauge: Ok, thanks!  The issue isn't the version of Powershell, but rather the AzureRM Powershell commandlets.  good advice though, let's combine it!  The command to get the major version of Azure powershell is:
(Get-Module -ListAvailable -Name Azure -Refresh).Version.Major

I'm updating this now.

petehauge:
Thanks for the review Leo - I've made all the suggested updates & tested the script with both Azure Powershell 5 and Azure Powershell 6, so we should be good to go!  NOTE:  I found that the Azure Powershell 6 commandlet still works with "ResourceName" (the old style, the new style is "Name") and now that I'm passing in the resource group name I don't need 2 different versions of the script.  Thanks!!

PR #392, opened 2018-07-19
Checking valid provisioning state and logging latest

leovms:
Modified wait logic for apply artifacts to check for valid provisioning state and log the latest value.

### DIFF
--- a/VstsTasks/AzureDtlCreateVM/task-funcs.ps1
+++ b/VstsTasks/AzureDtlCreateVM/task-funcs.ps1
@@ -32,7 +32,7 @@ function ConvertTo-Bool
     param(
         [string] $Value
     )
-    
+
     [bool] $boolValue = $false
 
     $null = [bool]::TryParse($Value, [ref]$boolValue)
@@ -46,7 +46,7 @@ function ConvertTo-Int
     param(
         [string] $Value
     )
-    
+
     [int] $intValue = 0
 
     $null = [int]::TryParse($Value, [ref]$intValue)
@@ -105,12 +105,12 @@ function Get-DeploymentTargetResourceId
         [string] $DeploymentName,
         [string] $ResourceGroupName
     )
-    
+
     [Array] $operations = Get-AzureRmResourceGroupDeploymentOperation -DeploymentName $DeploymentName -ResourceGroupName $ResourceGroupName
 
     foreach ($op in $operations)
     {
-        if ($op.properties.targetResource -ne $null)
+        if ($null -ne $op.properties.targetResource)
         {
             $targetResource = $op.properties.targetResource
             break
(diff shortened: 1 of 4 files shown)


VstsTasks/AzureDtlCreateVM/task-funcs.ps1
```diff
@@ -105,12 +105,12 @@ function Get-DeploymentTargetResourceId
...
         [string] $ResourceGroupName
     )
-    
+
     [Array] $operations = Get-AzureRmResourceGroupDeploymentOperation -DeploymentName $DeploymentName -ResourceGroupName $ResourceGroupName
 
     foreach ($op in $operations)
     {
-        if ($op.properties.targetResource -ne $null)
+        if ($null -ne $op.properties.targetResource)
```
michaelcohenmicrosoft: What if op.properties is null?

VstsTasks/AzureDtlCreateVM/task.json
```diff
@@ -12,7 +12,7 @@
     "version": {
         "Major": 2,
         "Minor": 0,
-        "Patch": 30
+        "Patch": 31
```
ajay-arora: doesn't it need to match with the version mentioned in vss-extension.json. if not, why are there 2 different versions.

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
```diff
@@ -414,9 +416,16 @@ function Wait-ApplyArtifacts
...
+
+            $continueWaiting = $true
+
             $vm = Get-AzureRmResource -ResourceId $ResourceId
-            $provisioningState = $vm.Properties.provisioningState
-            $continueWaiting = $provisioningState -eq 'ApplyingArtifacts' -or $provisioningState -eq 'UpgradingVmAgent'
+            if ($vm)
+            {
+                $provisioningState = $vm.Properties.provisioningState
+                $continueWaiting = $provisioningState -and ($provisioningState -eq 'ApplyingArtifacts' -or $provisioningState -eq 'UpgradingVmAgent')
```
ajay-arora: didnt we also discuss that we will reverse this check. by checking for succeeded/failed state instead. all other states continue waiting

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
leovms: The entire statement will be `null`. PS will not throw.

VstsTasks/AzureDtlCreateVM/task.json
leovms: The extension version is for the overall set of tasks. The task version is specific. Extension version needs to be incremented or it will not be updated in the Marketplace. The task version needs to be updated or the new scripts will not be used.

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
leovms: We did, but I don't think it is required. I've added the actual state in the VSTS logs. So, we'll be able to tell if that additional logic is really required.

PR #394, opened 2018-07-25
expose skipDos2Unix in the linux-run-bash artifact

lukka:
The linux-run-bash artifact does not expose the skipDos2Unix parameters, and its value is always 'false'. If you use the linux-run-bash artifact with a shell script that contains binary data appended at the end, you want to avoid to run the 'dos2unix' tool, and you want to avoid any manipulation of the downloaded file. It must be used as is, not conversion of its content at all.

Exposing this parameter allow to avoid any conversion of the downloaded script.


drwill-ms:
Confirmed with Luca that he's tested it. PR signed off.

PR #398, opened 2018-08-09
Fix for parameter names with blanks in environment templates

markusheiliger:


VstsTasks/vss-extension.json
```diff
@@ -1,7 +1,7 @@
 {
     "manifestVersion": 1,
     "id": "tasks",
-    "version": "1.0.44",
+    "version": "1.0.45",
```
leovms: Can you make this version `"1.0.46"`? Version 45 is already out there.

markusheiliger: sure

PR #407, opened 2018-09-08
Added artifact status to applying artifacts waiting logic.

leovms:


VstsTasks/AzureDtlCreateVM/task-funcs.ps1
```diff
@@ -417,15 +431,35 @@ function Wait-ApplyArtifacts
                 throw "Waited for more than $(ConvertTo-MinutesString $totalWaitMinutes). Failing the task."
             }
 
-            $continueWaiting = $true
-
             $vm = Get-AzureRmResource -ResourceId $ResourceId
```
drwill-ms: api version?

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
```diff
@@ -417,15 +431,35 @@ function Wait-ApplyArtifacts
...
+            }
+            else
+            {
+                $vmLabName = $(if ($vm.ParentResource){ $vm.ParentResource.Split('/')[-1] } else { $null })
+                $vmFullName = $(if ($vmLabName){ "$vmLabName/$vmName" } else { $vmName })
+            }
+            $vmResourceGroupName = $vm.ResourceGroupName
+            $vmResourceType = $vm.ResourceType
+
+            $vmDetails = Get-AzureRmResource -ApiVersion '2017-04-26-preview' -Name $vmFullName -ResourceGroupName $vmResourceGroupName -ResourceType $vmResourceType -ODataQuery '$expand=Properties($expand=Artifacts)'
```
drwill-ms: Can the expand to include artifacts not be used with the Get-AzureRmResource used with a resource Id?

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
```diff
@@ -417,15 +431,35 @@ function Wait-ApplyArtifacts
...
-                $provisioningState = $vm.Properties.provisioningState
-                $continueWaiting = $provisioningState -and ($provisioningState -eq 'ApplyingArtifacts' -or $provisioningState -eq 'UpgradingVmAgent')
+                throw "Unable to find VM with resource ID '$ResourceId'."
             }
 
+            $vmName = $vm.Name
+            if ($vm.ResourceName)
+            {
+                $vmLabName = $vm.ResourceName.Split('/')[0]
+                $vmFullName = $vm.ResourceName
```
drwill-ms: What is a VM *full* name?

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
leovms: No need to get the basic information for the resource. There are no expansions happening in this call.

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
leovms: `Api-Version` is not required to get the basic resource information. There are no expansions happening in this call.

VstsTasks/AzureDtlCreateVM/task-funcs.ps1
leovms: It contains the parent name. So, in the case of a Lab VM, it would be in the form `labName\vmName`. The resource name would only be `vmName`. But when we make the call to expand, we need the "full name". Otherwise, it doesn't work and it fails. Weird? Yes, but that's the change in the newer versions of the Azure PS cmdlet. 😞

PR #413, opened 2018-10-03
Azure DevOps Pipelines Agent - multiple instances in single DTL VM

onlyutkarsh:
Downloads latest Azure Pipelines agent for the account, and installs the specified number instances of the agents inside one DevTestLabs VM. Registers the agent with the Azure DevOps account, and adds agents to the specified agent pool. Note - If the specified agent folder already exists it will be overwritten.

Useful if you would like to install multiple agents inside a single VM.

image

### DIFF
--- a/Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
+++ b/Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
@@ -1,0 +1,53 @@
+{
+    "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2016-11-28/dtlArtifacts.json",
+    "title": "Azure Pipelines Multiple Agents",
+    "publisher": "onlyutkarsh",
+    "description": "Downloads latest Azure Pipelines agent, and creates the specified instances of the agents inside the specified VM. Registers the agent with the specified Azure DevOps account, and adds agents to the specified agent pool. Note - If the specified agent already exists it will be overwritten.",
+    "tags": [
+      "Azure DevOps",
+      "Pipelines",
+      "Release",
+      "Build",
+      "CI",
+      "Windows"
+    ],
+    "iconUri": "https://cdn.vsassets.io/content/icons/favicon.ico",
+    "targetOsType": "Windows",
+    "parameters": {
+      "Account": {
+        "type": "string",
+        "displayName": "Azure DevOps Account Name",
+        "description": "The name of the Azure DevOps account to add the build agent to. This is the value in your Azure DevOps URL: e.g. 'myaccount' in https://dev.azure.com/myaccount."
+      },
+      "PersonalAccessToken": {
+        "type": "securestring",
+        "displayName": "Azure DevOps Personal Access Token",
+        "description": "A personal access token with permissions to add build agents. It will only be used to register the agent."
+      },
+      "AgentName": {
+        "type": "string",
+        "displayName": "Agent Name",
+        "description": "The name to give to the agent, as seen by Azure DevOps. If empty, the computer name will be used. Prefixes every agent with given prefix",
+        "allowEmpty": true
+      },
+      "AgentNamePrefix": {
+        "type": "string",
+        "displayName": "Agent Name Prefix",
+        "description": "Prefixes every agent with string specified here",
+        "allowEmpty": true
+      },
+      "PoolName": {
+        "type": "string",
+        "displayName": "Agent Pool",
+        "description": "The agent pool this build agent should be added to."
+      },
+      "AgentCount":{
+        "type": "int",
+        "displayName": "Number of Agents",
+        "description": "The number of agents to install and add to the pool. Each agent will be named as prefix-agentname-(1..n)."
+      }
+    },
+    "runCommand": {
+      "commandToExecute": "[concat('powershell.exe -ExecutionPolicy bypass \"& ./run.ps1',' -Account ', parameters('Account'), ' -PersonalAccessToken ', parameters('PersonalAccessToken'), ' -AgentName ''', parameters('AgentName'), ''' -AgentNamePrefix ''', parameters('AgentNamePrefix'), ''' -PoolName ''', parameters('PoolName'), ''' -AgentCount ''', parameters('AgentCount'), '''\"')]"
+    }
+  }
--- a/Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
+++ b/Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
@@ -1,8 +1,8 @@
 {
     "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2016-11-28/dtlArtifacts.json",
-    "title": "Azure Pipelines Multiple Agents",
-    "publisher": "onlyutkarsh",
-    "description": "Downloads latest Azure Pipelines agent, and creates the specified instances of the agents inside the specified VM. Registers the agent with the specified Azure DevOps account, and adds agents to the specified agent pool. Note - If the specified agent already exists it will be overwritten.",
+    "title": "Azure Pipelines Agent(s)",
+    "publisher": "Utkarsh Shigihalli",
+    "description": "Downloads latest Azure Pipelines agent for the account, and installs the specified number instances of the agents inside one VM. Registers the agent with the Azure DevOps account, and adds agents to the specified agent pool. Note - If the specified agent folder already exists it will be overwritten.",
     "tags": [
       "Azure DevOps",
       "Pipelines",
--- a/Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
+++ b/Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
@@ -2,52 +2,65 @@
     "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2016-11-28/dtlArtifacts.json",
     "title": "Azure Pipelines Agent(s)",
     "publisher": "Utkarsh Shigihalli",
-    "description": "Downloads latest Azure Pipelines agent for the account, and installs the specified number instances of the agents inside one VM. Registers the agent with the Azure DevOps account, and adds agents to the specified agent pool. Note - If the specified agent folder already exists it will be overwritten.",
+    "description": "Downloads latest Azure Pipelines agent for the account, and installs the specified number instances of the agents inside one Windows VM. Registers the agent with the Azure DevOps account, and adds agents to the specified agent pool. Note - If the specified agent folder already exists it will be overwritten.",
     "tags": [
-      "Azure DevOps",
-      "Pipelines",
-      "Release",
-      "Build",
-      "CI",
-      "Windows"
+        "Azure DevOps",
+        "Pipelines",
+        "Release",
+        "Build",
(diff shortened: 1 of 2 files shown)


Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
```diff
@@ -0,0 +1,53 @@
+{
+    "$schema": "https://raw.githubusercontent.com/Azure/azure-devtestlab/master/schemas/2016-11-28/dtlArtifacts.json",
+    "title": "Azure Pipelines Agent(s)",
+    "publisher": "Utkarsh Shigihalli",
+    "description": "Downloads latest Azure Pipelines agent for the account, and installs the specified number instances of the agents inside one VM. Registers the agent with the Azure DevOps account, and adds agents to the specified agent pool. Note - If the specified agent folder already exists it will be overwritten.",
```
leovms: Consider making the overwrite of existing agents an option, since it is destructive. A user may or may intend to overwrite an existing agent. So, if the `$Overwrite` flag is set, then you can do it. Make sure that it's default is `False`. Then check for this condition inside your script and fail accordingly. Better make the user aware.

Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
```diff
@@ -0,0 +1,53 @@
...
+      },
+      "PersonalAccessToken": {
+        "type": "securestring",
+        "displayName": "Azure DevOps Personal Access Token",
+        "description": "A personal access token with permissions to add build agents. It will only be used to register the agent."
+      },
+      "AgentName": {
+        "type": "string",
+        "displayName": "Agent Name",
+        "description": "The name to give to the agent, as seen by Azure DevOps. If empty, the computer name will be used. Prefixes every agent with given prefix",
```
leovms: Is the last part of the description applicable here, in the form it is written? (i.e. `Prefixes every agent with given prefix`) It seems that if you remove that sentence, you still have complete details.

Artifacts/windows-azuredevops-multiple-agents/run.ps1
```diff
@@ -0,0 +1,122 @@
...
+(
+	[Parameter(Mandatory=$true)]
+	[string]$Account,
+
+	[Parameter(Mandatory=$true)]
+	[String]$PersonalAccessToken,
+
+	[string]$AgentName = $env:COMPUTERNAME,
+
+	[string]$AgentNamePrefix= "dtl",
```
leovms: Remove default, set it in the `artifactfile.json` instead, so it will also show in the UI to the user.

Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
```diff
@@ -0,0 +1,53 @@
...
+        "displayName": "Agent Name Prefix",
+        "description": "Prefixes every agent with string specified here",
+        "allowEmpty": true
+      },
+      "PoolName": {
+        "type": "string",
+        "displayName": "Agent Pool",
+        "description": "The agent pool this build agent should be added to."
+      },
+      "AgentCount":{
```
leovms: Consider setting the default to `1` and make this mandatory to be more explicit to the user. I do see in the PS1 logic that we do `$AgentCount + 1` in the `for` loop, which will give `1` if not set. So, the logic will still work. Food for thought.

Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
```diff
@@ -0,0 +1,53 @@
...
+        "type": "string",
+        "displayName": "Azure DevOps Account Name",
+        "description": "The name of the Azure DevOps account to add the build agent to. This is the value in your Azure DevOps URL: e.g. 'myaccount' in https://dev.azure.com/myaccount."
+      },
+      "PersonalAccessToken": {
+        "type": "securestring",
+        "displayName": "Azure DevOps Personal Access Token",
+        "description": "A personal access token with permissions to add build agents. It will only be used to register the agent."
+      },
+      "AgentName": {
```
leovms: Are agent names with spaces allowed? Can you confirm that? Maybe there's some additional logic you may need to add, if that's not the case.

Artifacts/windows-azuredevops-multiple-agents/run.ps1
```diff
@@ -0,0 +1,122 @@
+# Downloads the Azure DevOps Pipelines Agent and installs specified instances on the new machine
+# under C:\agents\ and registers with the Azure DevOps Pipelines agent pool
+
+# Enable -Verbose option
+[CmdletBinding()]
+Param
+(
+	[Parameter(Mandatory=$true)]
```
leovms: Don't make these `[Parameter(Mandatory=$true)]`. Just validate within your script and `throw` accordingly. The reason is that if a user mistakenly passes no values, without knowing and if allowed, then the script will hang, waiting for input.

Artifacts/windows-azuredevops-multiple-agents/run.ps1
```diff
@@ -0,0 +1,122 @@
...
+    $basicAuth = [System.Text.Encoding]::UTF8.GetBytes($basicAuth)
+    $basicAuth = [System.Convert]::ToBase64String($basicAuth)
+    $headers = @{ Authorization = ("Basic {0}" -f $basicAuth) }
+    $agentList = Invoke-WebRequest -Uri $vstsAgentUrl -Headers $headers -Method Get -ContentType application/json -UseBasicParsing | ConvertFrom-Json
+    $agent = $agentList.value
+    if ($agent -is [Array])
+    {
+        $agent = $agentList.value[0]
+    }
+    $agentPackagePath = "$agentTempFolderName\agent.zip"
```
leovms: If the file is partially created because the download fails, and we retry, will the existence of the file cause the subsequent call to download to fail or will the file be overwritten? From the logic, the temp folder would be the same when retrying. So, you may want to confirm. Just add a dummy `agent.zip` file in the folder and see if the call to download still works and overwrites as expected. Otherwise, in your `catch`, you may need to delete the `agent.zip` file before retrying.

Artifacts/windows-azuredevops-multiple-agents/run.ps1
```diff
@@ -0,0 +1,122 @@
...
+	# Don't recurse down _work or _diag, those files are not blocked and cause the process to take much longer
+	Write-Verbose "Unblocking files" -verbose
+	Get-ChildItem -Recurse -Path $agentInstallationPath | Unblock-File | out-null
+
+	# Retrieve the path to the config.cmd file.
+	$agentConfigPath = [System.IO.Path]::Combine($agentInstallationPath, 'config.cmd')
+	Write-Verbose "Agent Location = $agentConfigPath" -Verbose
+	if (![System.IO.File]::Exists($agentConfigPath))
+	{
+		Write-Error "File not found: $agentConfigPath" -Verbose
```
leovms: `Write-Error` will only be terminal if you `$ErrorActionPreference = "Stop"`. So, instead, use `throw` here and `return` is redundant, so you can remove it.

Artifacts/windows-azuredevops-multiple-agents/run.ps1
```diff
@@ -0,0 +1,122 @@
...
+		return
+	}
+
+	# Call the agent with the configure command and all the options (this creates the settings file) without prompting
+	# the user or blocking the cmd execution
+	Write-Verbose "Configuring agent '$($Agent)'" -Verbose		
+	.\config.cmd --unattended --url $serverUrl --auth PAT --token $PersonalAccessToken --pool $PoolName --agent $Agent --runasservice
```
leovms: You may need to test the last exit code after calling `.\config.cmd` and throw to ensure that the script fails correctly. Here's a helper function you can use for that:

``` PS
function Test-LastExitCode
{
    param
    (
        [string] $Message = ''
    )

    # Whenever we execute commands such as '& somecommand' or via Invoke-Expression, we should always
    # check the exit code, so we can decide to stop processing at that time, by throwing an exception.
    $exitCode = $LASTEXITCODE
    if ($exitCode -and $exitCode -ne 0)
    {
        if ($Message)
        {
            if (-not $Message.EndsWith('.'))
            {
                $Message += '.'
            }
            $Message += ' '
        }
        $Message += "Last command exited with error code $exitCode"
        throw $Message
    }
}
```
So, you would use it like this:
``` PS
.\config.cmd ...
Test-LastExitCode
```

Artifacts/windows-azuredevops-multiple-agents/run.ps1
```diff
@@ -0,0 +1,122 @@
...
+	
+	Write-Verbose "Agent install output: $LASTEXITCODE" -Verbose
```
leovms: If you use the call to `Test-LastExitCode`, it will already document the exit code for you.

Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
```diff
@@ -0,0 +1,53 @@
...
+        "type": "string",
+        "displayName": "Agent Name",
+        "description": "The name to give to the agent, as seen by Azure DevOps. If empty, the computer name will be used. Prefixes every agent with given prefix",
+        "allowEmpty": true
+      },
+      "AgentNamePrefix": {
```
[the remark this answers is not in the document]
onlyutkarsh: Hi, @leovms, can you please elaborate on this one, please? I am removing `allowEmpty: true` for `AgentNamePrefix` parameter. Is there something else required?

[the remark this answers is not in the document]
onlyutkarsh: Okay, ignore that. Found that you can set default value like this `"defaultValue": 0`. I was looking [here](https://github.com/Azure/azure-devtestlab/blob/master/Documentation/AUTHORING.md#parameters) which probably needs update.

Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
onlyutkarsh: Agree. Will make the change

Artifacts/windows-azuredevops-multiple-agents/artifactfile.json
onlyutkarsh: You can create agents with space. Let me know if there is any change required.

onlyutkarsh:
@leovms Thanks for the review and your detailed comments. I have made the changes to the files as per your comments. Please take a look.

Artifacts/windows-azuredevops-multiple-agents/run.ps1
```diff
@@ -0,0 +1,122 @@
...
+$retryCount = 3
+$retries = 1
+Write-Verbose "Downloading Agent install files" -verbose
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+do
+{
+  try
+  {
+    Write-Verbose "Trying to get download URL for latest VSTS agent release..."
+    $vstsAgentUrl = "$serverUrl/_apis/distributedtask/packages/agent/win7-x64?`$top=1&api-version=3.0"
```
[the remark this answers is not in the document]
onlyutkarsh: fixed it in my last commit

onlyutkarsh:
@leovms let me know if any further changes required.

PR #414, opened 2018-10-04
Removed references to obsolete Find-AzureRmResource cmdlet

leovms:
- Split Classic and Resource Manager scripts
- Updated logic to remove reference to obsolete Find-AzureRmResource cmdlet.

### DIFF
--- a/Scripts/CopyClassicVHDFromVMToLab.ps1
+++ b/Scripts/CopyClassicVHDFromVMToLab.ps1
@@ -1,0 +1,253 @@
+[CmdletBinding()]
+Param(
+    # Enter the subscription ID. It is assumed that both source VM and destination lab are both in
+    # this subscription.
+    [Parameter(Mandatory)]
+    [string]
+    $SubscriptionId = '<SubscriptionId>',
+
+    # Enter the name of the Lab where you want to copy the VHD file.
+    [Parameter(Mandatory)]
+    [string]
+    $LabName = '<LabName>',
+    
+    # Enter the resource group name of the Lab where you want to copy the VHD file.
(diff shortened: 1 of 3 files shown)


Scripts/CopyClassicVHDFromVMToLab.ps1
```diff
@@ -0,0 +1,253 @@
...
+
+###################################################################################################
+#
+# Main execution block.
+#
+
+try
+{
+    Write-Host "Selecting subscription '$SubscriptionId' ... " -NoNewline
+    Select-AzureSubscription -SubscriptionId $SubscriptionId -ErrorAction SilentlyContinue | Out-Null
```
drwill-ms: should we silently continue?

leovms: We used to need it when we had both scripts in one. Now that they are separate, I can change it. In this case, it makes sense that both are correctly set.

Scripts/CopyClassicVHDFromVMToLab.ps1
```diff
@@ -0,0 +1,253 @@
...
+    [string]
+    $VMServiceName = '',
+
+    # Enter the name of the VHD file with extension as .vhd. You will identify the file with this
+    # name while creating template.
+    [Parameter(Mandatory)]
+    [string]
+    $VHDFileName = '<VHDFileName>.vhd',
+    
+    # Enter the seconds after the Shared Access Signature on source will expired. Default value ist 3600.
```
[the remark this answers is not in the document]
leovms: Thanks. I'll fix it.

PR #430, opened 2019-01-31
New ARM template to deploy SharePoint 2013/2016/2019

Yvand:
I propose this ARM template, that is largely based on [my other template](https://azure.microsoft.com/en-us/resources/templates/sharepoint-adfs/), to replace the existing SharePoint template.
This one has many improvements compared to the current one, for example:

- User can choose to deploy SharePoint 2013, 2016, 2019 individually, or all simultaneously
- Template uses [SharePointDsc](https://github.com/PowerShell/SharePointDsc), as well as all other recent DSC modules needed to deploy the environment
- It uses relevant/recent Azure features (for example, Network Security Groups and Managed disks)
- Deployment is optimized for running integration tests: minimum configuration to provision quickly, but still viable to cover most scenarios.

PR #450, opened 2019-03-19
Added the SharePoint 2019 StandAlone Dev Environment ARM Template

NikCharlebois:


NikCharlebois:
Hold on to this PR. We need to straighten our approach for SP based on other templates available first.

leovms:
Changes are being incorporated as part of #430.

PR #456, opened 2019-04-04
Add samples for Lab Services dotnet SDK

ianhays:
Adds two samples for LabServices:
1. Create RDP files for all VMs owned by a user across all labs
2. Create RDP files for all VMs under a lab

PTAL: @leovms


Samples/LabServices/DotnetSDK/Instructions.txt
```diff
@@ -0,0 +1,16 @@
+How to set up a simple app to use the .NET SDK for Azure Lab Services
```
leovms: Make this a `README.md` and it should get picked up as the folder's documentation. Use formatting accordingly.

ianhays:
> Under `Samples` , should we be calling the folder `LabServices` or 'ClassroomLabs`?

I was thinking yes originally, but the change of naming to `ClassroomLabs` has not been done yet in any public capacity. So this would be the only place where we call it `ClassroomLabs`, which would be confusing. 

I think we should leave the name here as `LabServices` to be consistent with the naming of the package ([Microsoft.Azure.Management.LabServices](https://www.nuget.org/packages/Microsoft.Azure.Management.LabServices/)) as well as the naming [in the documentation](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/labservices?view=azure-dotnet) until the official name change happens everywhere.

PR #457, opened 2019-04-08
Add install powershell module artifact

ross-bohr:
New DevTest Labs artifact added: windows-install-powershell-module. Uses a custom ps script to run Install-Module and Import-Module cmdlets.
Tested on Windows Server 2019.


Artifacts/windows-install-powershell-module/install-powershell-module.ps1
```diff
@@ -0,0 +1,94 @@
...
+
+#
+# Main execution block.
+#
+
+try
+{
+    Write-Host "Installing module $moduleName"
+
+    If(-NOT ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator"))
```
leovms: Why is this required? The execution context should be with sufficient permissions. Have you experienced issues running this script as an artifact that would require this part of the logic?

Artifacts/windows-install-powershell-module/Artifactfile.json
```diff
@@ -0,0 +1,25 @@
+{
```
leovms: Rename artifact folder to `windows-powershellmodule` to follow the convention used for other PS artifacts.

Artifacts/windows-install-powershell-module/install-powershell-module.ps1
```diff
@@ -0,0 +1,94 @@
...
+        # Start the new process
+        Start-Process powershell.exe -Verb runas -ArgumentList $arguments
+        exit
+    }
+    else
+    {
+		Install-Module $moduleName -AllowClobber -Force -Verbose
+		Write-Host "Installed module $moduleName"
+
+		Import-Module $moduleName -Verbose
```
leovms: Why are we using `Import-Module` here? Does it not just apply to the current session? Is it being used to validate the installation of the module?

Artifacts/windows-install-powershell-module/install-powershell-module.ps1
```diff
@@ -0,0 +1,94 @@
...
+#
+
+try
+{
+    Write-Host "Installing module $moduleName"
+
+    If(-NOT ([Security.Principal.WindowsPrincipal] [Security.Principal.WindowsIdentity]::GetCurrent()).IsInRole([Security.Principal.WindowsBuiltInRole] "Administrator"))
+    {
+        #build up the deploy arguments
+        $arguments = "-file `"{0}`"" -f $script:MyInvocation.MyCommand.Path
```
leovms: How are we passing the `$moduleName`? We should also refrain from using `-File`. Look at the `artifactfile.json` and check `commandToExecute`. We should use that format instead to bubble up errors.

Artifacts/windows-install-powershell-module/install-powershell-module.ps1
```diff
@@ -0,0 +1,94 @@
...
+    
+        # Start the new process
+        Start-Process powershell.exe -Verb runas -ArgumentList $arguments
+        exit
+    }
+    else
+    {
+		Install-Module $moduleName -AllowClobber -Force -Verbose
```
leovms: On Windows 10, there are dependencies missing, for example: `PowerShellGet requires NuGet provider version '2.8.5.201' or newer to interact with NuGet-based repositories.` Therefore, the logic will not work for all PS versions and OS flavors.

leovms:
@ross-bohr, I've modified this artifact in my own private repo. I'd be happy to update your PR with my changes. I've made the necessary changes to make it run correctly in Windows 10. I've also updated the definition to indicate that PS 5+ is required for this module to work. We should consider adding support for earlier versions of PS, as they are still wildly used and such an artifact can be useful for those versions as well.

leovms:
I went ahead and applied my changes over yours. Feel free to let me know if you have any objections. I'm going ahead and pushing the changes to the Public Repo.

PR #487, opened 2019-05-17
Change Delete Environment to log message when environment not found

RogerBestMsft:
Throwing an error on delete environment when the environment to delete doesn't exist is an annoyance.  Changing it to log a message.


leovms:
Thank you for making the updates. The extension has been published. When the Delete Environment task is used, it should display version 1.0.5.

PR #493, opened 2019-05-29
DomainAdminPassword Should Be SecureString Param Type

michaelwstark:


michaelwstark:
@leovms - Do we have a setup to test this change?

Issue #494, opened 2019-05-30
[AZ.DEVTESTLABS]  Library doesn't work on Azure Powershell 5.x


petehauge:
When using the DevTest Labs azure module, even running basic commands like "Dtl-GetLab" fails.  
![libraryerror](https://user-images.githubusercontent.com/12090616/58667221-70ac6e80-8303-11e9-9216-dbb0f8d8bc45.png)

The underlying cause is approx. line 80:
    Get-AzureRmResource -ExpandProperties -resourcetype $ResourceType -EA SilentlyContinue | Where-Object { $_.ResourceId -like "*/resourcegroups/$ResourceGroupName/*/labs/$Name"}

Unfortunately calling Get-AzureRmResource with only a resource type wasn't supported until a later version of powershell (once they got rid of the Find-AzureRmResource commandlet).

We need to decide if we require Azure Powershell 6.x or newer (or Azure Powershell Core 1.x or newer), or if we should make the library compatible with Azure Powershell 5.x...

lucabol:
Fixed in #496 by adding test for version. We don't work in 4.X.

Issue #495, opened 2019-05-30
[AZ.DEVTESTLABS] Library doesn't allow querying for VMs in a "Failed" state


petehauge:
![libraryerror](https://user-images.githubusercontent.com/12090616/58667318-ac473880-8303-11e9-9c22-212b5fc6fb34.png)

The DTL PowerShell library currently doesn't support querying for VMs in a **'failed'** state, we should add this.

lucabol:
Fixed in #496

PR #496, opened 2019-05-30
Update to a more reliable way to get VM Status

petehauge:
Fix for https://github.com/Azure/azure-devtestlab/issues/495

Updating the code for getting the status of a VM to jump to the underlying Compute VM for the 'real' status.  In most cases, this is 1 additional call to Azure (Get-AzureRmVm -Status)


petehauge:
Also adding code to fix https://github.com/Azure/azure-devtestlab/issues/494 , we are going to block on older versions of Azure PowerShell.

image

lucabol:
Suggest putting code under '-ExtendedState' parameter for performance reason and add an -'ExtendedState' parameter to dtl-getvm as well

PR #497, opened 2019-05-31
Add Remove Appx packages artifact

azabbasi:


Artifacts/windows-appxremove/remove-appx-packages.ps1
```diff
@@ -0,0 +1,60 @@
...
+    # a non-zero exit code from the catch block.
+    exit -1
+}
+
+###################################################################################################
+#
+# Main execution block.
+#
+
+try
```
leovms: `try` is not required, unless you have a `catch` or `finally` block. The `trap` will handle the error for you.

Artifacts/windows-appxremove/remove-appx-packages.ps1
```diff
@@ -0,0 +1,60 @@
...
+{
+    if ([Environment]::OSVersion.Version.Major -gt 6) {
+        Write-Host "`nRemoving Appx Packages`n"
+        Get-AppxPackage -AllUsers | Remove-AppxPackage -ErrorAction SilentlyContinue
+        Write-Host "`nDone Removing Appx Packages`n"
```
leovms: Not sure if you need to use \`n for in all your strings. You will end up with blank lines between each. If that's on purpose, ignore my comment.

Artifacts/windows-appxremove/remove-appx-packages.ps1
```diff
@@ -0,0 +1,60 @@
...
+###################################################################################################
+#
+# Main execution block.
+#
+
+try
+{
+    if ([Environment]::OSVersion.Version.Major -gt 6) {
+        Write-Host "`nRemoving Appx Packages`n"
+        Get-AppxPackage -AllUsers | Remove-AppxPackage -ErrorAction SilentlyContinue
```
leovms: If you call `Remove-AppxPackage` with `-ErrorAction SilentlyContinue`, then you are potentially never going to fail.

Artifacts/windows-appxremove/remove-appx-packages.ps1
azabbasi: done

Artifacts/windows-appxremove/remove-appx-packages.ps1
azabbasi: done

PR #499, opened 2019-06-01
Adding version check for AzureRm

lucabol:


lucabol:
https://github.com/Azure/azure-devtestlab/pull/496 already fixes it.

PR #503, opened 2019-06-11
Adding support for selecting 32-bit vs 64-bit of VS Code

leovms:


Artifacts/windows-vscode/Artifactfile.json
```diff
@@ -2,15 +2,28 @@
...
+      "Windows",
+      "VSCode",
+      "Visual Studio"
     ],
     "iconUri": "https://github.com/Azure/azure-devtestlab/raw/master/Artifacts/windows-vscode/vscode.png",
     "targetOsType": "Windows",
+    "parameters": {
+      "Version": {
+        "type": "string",
+        "displayName": "Version",
```
michaelwstark: Maybe Architecture would be a better name.  I think Version would be confused for X.X.X.X strings.

Artifacts/windows-vscode/Artifactfile.json
```diff
@@ -2,15 +2,28 @@
...
+        "description": "Version of Visual Studio Code to install. Valid values are 32-bit or 64-bit.",
+        "defaultValue": "32-bit",
+        "allowEmpty": false,
+        "allowedValues": [
+          "32-bit",
```
michaelwstark: Would Default be a good option to have?  Essentially, x64 unless the machine is x86 only?  If someone wants a specific version they can choose a specific value?

Artifacts/windows-vscode/Artifactfile.json
leovms: Yes, it was considered. It just didn't look right, but we can definitely change it.

Artifacts/windows-vscode/Artifactfile.json
leovms: Yes, this was done for backward compatibility. If the artifact is being used in a formula, adding the parameter without a default would break it. So, the safest thing is to use the default. I do have a version of the artifact with "Auto" as an option, in addition to the 32-bit and 64-bit. But again, because of backward compatibility, this is the better approach, I think.

Artifacts/windows-vscode/Artifactfile.json
PhilLee-MSFT: I'd guess that all the folks using this will know about architecture.

rgrwatson85:
@leovms @michaelwstark 

This is a breaking change for those that are using this as a mandatory artifact. To make this backwards compatible, it could have been implemented as two distinct artifacts that point at the different architectures.

leovms:
@rgrwatson85, thank you for reporting this. Let me take a look at this right away. I’ll double check and prepare a fix right away. Can you tell me what is the error you see?

rgrwatson85:
@leovms - mandatory artifacts cannot have parameters

` "json": {
        "error": {
            "code": "ArtifactNotParameterless",
            "message": "The artifact '/subscriptions/3fe02f66-a68a-4cc3-bf2e-625e5be45301/resourceGroups/dc-demo-lab/providers/Microsoft.DevTestLab/labs/dc-demo-lab/artifactSources/public repo/artifacts/windows-vscode' requires 1 parameters, but only artifacts without parameters are allowed as mandatory artifacts."
        }
    }`

rgrwatson85:
@leovms - mandatory artifacts cannot have parameters
"json": { "error": { "code": "ArtifactNotParameterless", "message": "The artifact '/subscriptions/……./resourceGroups/……./providers/Microsoft.DevTestLab/labs/...…./artifactSources/public repo/artifacts/windows-vscode' requires 1 parameters, but only artifacts without parameters are allowed as mandatory artifacts." } }

leovms:
I see. I did miss that. I will revert the change and create a new artifact for that. Thank you.

EricCote:
@leovms You talk about creating a new artifact for VsCode 64. But here we are 2 years later, and the VsCode artifact is still 32-bit.  

I could create a pull request for VsCode 64 bits.  Should I do that?

leovms:
Yes, please. This got de-prioritized. So, feel free to contribute. Much appreciated.

PR #505, opened 2019-06-13
Mini-harness for running Pester in Azure Pipelines and fix tests

petehauge:


lucabol:
Fill free to simply check-in the tests without doing a PR. We will review the whole thing together when we get the budge on the Readme.

PR #509, opened 2019-06-14
Updating scenario tests for correct powershell syntax

petehauge:


lucabol:
Thx for doing this. Another piece of the puzzle gets into place.

PR #510, opened 2019-06-14
Adding logic to ensure unique arguments are passed to installer

leovms:
- Changed logic used to contructs arguments.
- Added logic to remove duplicate arguments.
- Consolidated into a single PS1.
- Added default values to artifact so UI is not blank at first load.


Artifacts/windows-visualstudio/artifact.ps1
```diff
@@ -43,78 +90,133 @@ function DownloadToFilePath ($downloadUrl, $targetFile)
...
+        }
+
+        throw "Product installation of $localFile failed with exit code: $($retCode.ExitCode.ToString())"    
     }
 
-    Write-Error "Product installation of $localFile failed with exit code: $($retCode.ExitCode.ToString())"    
-}
-else
-{
     Write-Output "Visual Studio install succeeded. Rebooting..."
```
joerage: Earlier, you add the option -norestart. So this would not be rebooting, right?

leovms: The logic is meant to allow control of the rebooting by the artifact and not the VS installer. So, in the artifact definition, we do have a restart post deploy action.

PR #518, opened 2019-06-20
Finalize CI pipeline for DTL Library Module

petehauge:
**Three changes included:**

* Ensure that all the names we use when deploying Azure resources are random.  We think we were seeing collisions that was causing test failures
* Add the pipeline status badge to the readmes for the DTL Module (setup to run nightly)
* Fix a few places on the "Should" commandlet where we didn't use the right syntax


lucabol:
I will let @leovms decide to add the badge to the overall DTL Readme. We could do it later after we make sure that it is green most times :-)

README.md
```diff
@@ -15,6 +15,10 @@ This is a public, community-contributed repository that contains:
 
 [Az.DevTestLabs](https://github.com/Azure/azure-devtestlab/tree/master/samples/DevTestLabs/Modules/Library), for short, is a PowerShell module to simplify the management of [Azure DevTest Labs](https://azure.microsoft.com/en-us/services/devtest-lab/). It provides composable functions to create, query, update and delete labs, VMs, Custom Images and Environments. The source can be located under [/samples/DevTestLabs/Modules/Library/](https://github.com/Azure/azure-devtestlab/tree/master/samples/DevTestLabs/Modules/Library).
 
+**PowerShell Module Status:**
```
leovms: Not sure this is required here. You already have it in the library `README.md`, to which the above text already links.

petehauge: No problem - @leovms , I can get rid of the status here now.

petehauge: @leovms - done!  I got rid of this from the main readme.

PR #520, opened 2019-06-24
Add in "Get-AzDtlVmRdpFileContents" and tests

petehauge:
Based on a customer request, they needed a PowerShell script to get the RDP File Contents for a particular VM, so we added the code directly to the library for everyone along with accompanying tests.

I also fixed the code for getting the header bearer token, it didn't work when there was only Az commandlets (no AzureRM) on the machine.


lucabol:
Looks good to me. Thanks for the Headers code. I'll add it to the azlabs library as well. It's a tricky matter to get the Az vs AzureRm code right.

PR #530, opened 2019-07-12
Fix: Installation may run into timeout 

rzander:
Sub-Component installation did wait forever


Artifacts/windows-ConfigurationManagerCurrentBranch(EVAL)/setupCMCB.ps1
```diff
@@ -60,7 +60,7 @@ if((gwmi win32_computersystem).partofdomain -eq $true)
     Start-Service 'MSSQLSERVER'
 
 	#Install ADK10
-	$proc = (Start-Process -FilePath "$($env:temp)\RZUpdate.exe" -ArgumentList "ADK10" -Wait -PassThru)
```
leovms: The original author may have wanted to wait for the process to complete for a reason. I wonder if removing the parameter will have undesirable side effects.

Would it not be better to use Wait-Process? Something like the following:
```PS1
try
{
    $proc | Wait-Process -Timeout $MaxWait -ErrorAction Stop
}
catch
{
    $proc | Stop-Process -Force
}
```

rzander: ```Start-Process -wait``` does wait for the specified process and its **descendants** to complete (https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/start-process?view=powershell-6).  
There is still a ```$proc.WaitForExit()``` which does wait until the main process is finished. 

Additionally, there are some checks in the script which will exit with an error code if something is wrong (1 = SQL missing; 2 = ADK missing; 3 ConfigMgr missing)

On slow machines i've seen that the artifact failed but the script was still running, so I assume that artifacts do also have a timeout ?! On a slow Machine, this installation can take up to two hours...

leovms: Applying artifacts times out after 2 hours. What are we really fixing then? Sorry, I missed the original issue. Is the code change you're making allowing the logic to fail sooner than the artifact time out period?

rzander: The "-wait" does wait forever on the "ConfigMgr" installation because the setup will generate some child processes which do not end... 
Removing the "-wait" does guarantee that the installation continues when rzupdate.exe is finished.... independent of the artifact timeout.

It's just the "-wait" on "start-process", everything else is fine...

leovms: Fair enough. I don't know enough about the artifact. I'll assume you validated all this. So, I'll go ahead and approve it.

PR #546, opened 2019-08-06
Fix Get AzureRM major version

remote-specialist:
Get-Module -Name "AzureRM" -ListAvailable | Sort-Object Version.Major -Descending | Select-Object -First 1

this code will return version 5 if version 5 and 6 are installed

### DIFF
--- a/samples/DevTestLabs/Modules/Library/Az.DevTestLabs2.psm1
+++ b/samples/DevTestLabs/Modules/Library/Az.DevTestLabs2.psm1
@@ -12,7 +12,7 @@ Set-StrictMode -Version Latest
 # If you have the AzureRm module, then everything works fine
 # If you have the Az module, we need to enable the AzureRmAliases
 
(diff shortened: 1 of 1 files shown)


petehauge:
Thanks for looking into this!  I installed multiple versions of Azure RM including 5 + 6 and ran the original code, and it works as expected.  The new code also works for this case.

Unfortunately, if there is NO AzureRM module installed there's an error with the new code..

azurermversion


azurermversionerror

remote-specialist:
For me, it does not work. In my case, I got the same results with and without -Descending flag
screen1

petehauge:
How about we do something like this?  It's your code with ignoring the error so we don't end up with a failure if the user doesn't have AzureRM installed at all:

Get-InstalledModule -Name "AzureRM" -ErrorAction SilentlyContinue

remote-specialist:
For me, such a solution is perfectly fine

PR #556, opened 2019-08-13
Az.LabsServices: fix exception when retrieving not existing lab account

lucabol:


lucabol:
Merging it in so can move forward with coding :-) Could have simply checked it in as very small. I think I messed up my git process, as shouldn't find myself in this situation.

PR #559, opened 2019-08-14
Az.LabServices: create fast tests and refactor common tasks to utils.ps1

lucabol:
This puts in place the new test strategy with the following goals:
1. Use hosted agent pool instead of custom.
2. Have set of Fast tests to run on demand when working on features.
3. Run all tests nightly to catch regressions

Changes to code:
1. Rename tests with Fast prefix
2. Have them work on an existing lab account and lab
3. Refactor common functions in utils.ps1
4. Change Readme to take badge from All Tests nighly pipeline on master

Changes to pipeline:
1. Create two hosted agent pipeline for Fast and All tests
2. Fast runs all that start with Fast*, All runs all
3. All runs nightly, Fast on demand


samples/ClassroomLabs/Modules/Library/Tests/LabAccount.test.ps1
```diff
@@ -0,0 +1,30 @@
...
+            
+            $la | Should -Not -BeNullOrEmpty
+        }
+
+        It 'Can remove Lab Account' {
+
+            $la = Get-AzLabAccount -ResourceGroupName $rgName -LabAccountName $laName
+            $la | Remove-AzLabAccount
+
+            Get-AzLabAccount -ResourceGroupName $rgName -LabAccountName $laName | Should -Not -BeNullOrEmpty
```
petehauge: The lab account was just removed - shouldn't we validate "Should -BeNullOrEmpty" instead of "Should -Not -BeNullOrEmpty"?

samples/ClassroomLabs/Modules/Library/Tests/Lab.tests.ps1
```diff
@@ -0,0 +1,74 @@
+[cmdletbinding()]
+Param()
+Import-Module $PSScriptRoot\..\Az.LabServices.psm1
+
+$rgName = 'TestPerm'
```
petehauge: Is there any way we can prepend the resource names (at least RGs and Lab Accounts) with something across all the tests?  That way I know what resources in the subscription are connected to the automation?  On the other library I put "DtlLibrary…" at the beginning...

samples/ClassroomLabs/Modules/Library/Tests/Utils.ps1
```diff
@@ -0,0 +1,63 @@
...
+$imgName = 'CentOS-Based*'
+$maxUsers = 2
+$usageQuota = 30
+$usageAMode = 'Restricted'
+$shPsswd = $false
+$size = 'Medium'
+$title = 'Advancing Differentiation Workshop'
+$descr = 'Bringing it to the 21st Century'
+$userName = 'test0000'
+$password = 'Test000'
```
petehauge: I wonder if we should have another way to handle credentials, like storing in keyvault and reading them?  Or generating and saving them somewhere?  Can do that in the future...

samples/ClassroomLabs/Modules/Library/Tests/Utils.ps1
```diff
@@ -0,0 +1,63 @@
...
+
+function Get-FastLab {
+    [CmdletBinding()]
+    param()
+
+    $la = Get-FastLabAccount
+                
+    $lab = $la | Get-AzLab -LabName $labName
+    if ($lab) {
+        return $lab
```
petehauge: Do we need to ensure the lab account is in any particular state?  I see that there was some handling for this in one of the tests, but it might be better to reset to a baseline here?  (can be done later, not required for this PR)

samples/ClassroomLabs/Modules/Library/Tests/Fast-SharedGallery.test.ps1
```diff
@@ -0,0 +1,28 @@
...
+. $PSScriptRoot\Utils.ps1
+
+$la = Get-FastLabAccount
+$sg = Get-AzResource -resourceId "/subscriptions/39df6a21-006d-4800-a958-2280925030cb/resourceGroups/SharedImageGalleryRG/providers/Microsoft.Compute/galleries/EnterpriseSharedImages"
+
+Describe 'Shared Gallery Management' {
+    It 'Can attach a shared library' {
+            
+        # to run locally uncomment the line below and comment the next one
+        #$sg = Get-AzResource -resourceId "/subscriptions/d5e481ac-7346-47dc-9557-f405e1b3dcb0/resourceGroups/myGalleryRG/providers/Microsoft.Compute/galleries/myGallery"
```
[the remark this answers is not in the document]
lucabol: Done.

samples/ClassroomLabs/Modules/Library/Tests/Fast-SharedGallery.test.ps1
```diff
@@ -0,0 +1,28 @@
+[cmdletbinding()]
+Param()
+Import-Module $PSScriptRoot\..\Az.LabServices.psm1
+
+. $PSScriptRoot\Utils.ps1
+
+$la = Get-FastLabAccount
+$sg = Get-AzResource -resourceId "/subscriptions/39df6a21-006d-4800-a958-2280925030cb/resourceGroups/SharedImageGalleryRG/providers/Microsoft.Compute/galleries/EnterpriseSharedImages"
```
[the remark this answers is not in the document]
lucabol: I went for the last idea you suggested. I give an error if don't find one, but could change to create one if needed

samples/ClassroomLabs/Modules/Library/Tests/LabAccount.test.ps1
lucabol: My intuition is that it was working before because it would get the lab account before the removal was complete. I can either make the removal wait for completion or remove the check. I went for the latter as making the removal synchronous is not trivial and perhaps not correct either (aka in practical code, why wait for it?). In any case, if you consider it wrong, reopen the conversation.

samples/ClassroomLabs/Modules/Library/Tests/Lab.tests.ps1
lucabol: Done. AzLabsLibrary is the prefix.

samples/ClassroomLabs/Modules/Library/Tests/Utils.ps1
lucabol: Yes! I want to be able to do multiple runs in parallel accessing the same lab. The tests as written now don't allow this. What you suggest doesn't allow it either as, while the baseline is resetting, another test run might be changing the lab. It needs to be a mix of writing the test in an atomic way and judicious use of mutex when nothing else works. I have ideas on all of that, but they are worth a separate PR.

samples/ClassroomLabs/Modules/Library/Tests/Utils.ps1
lucabol: Yep, it does look bad but rather harmless as written. Agree to revisit in the future.

PR #562, opened 2019-08-20
Implementing Azure DevTest Labs Tasks in Node.js

leovms:


tasks/Node/webpack.config.js
```diff
@@ -0,0 +1,29 @@
...
+    AzureDtlDeleteEnvironment: "./src/tasks/AzureDtlDeleteEnvironment/task.ts",
+    AzureDtlDeleteVM: "./src/tasks/AzureDtlDeleteVM/task.ts",
+    AzureDtlUpdateEnvironment: "./src/tasks/AzureDtlUpdateEnvironment/task.ts"
+  },
+  output: {
+    path: __dirname + "/dist/",
+    filename: "tasks/[name]/task.js"
+  },
+  resolve: {
+    extensions: [".ts", ".tsx", ".js"]
```
mharlan: We don't need the .tsx extension, but there isn't any harm in resolving it.

tasks/Node/webpack.config.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
+//  mode: "development",
```
mharlan: Do you want to clean up the commented out mode an devtool settings or did you want to leave them around?

leovms: I figured we could leave it for the next person who may do some work and is just learning about Webpack. But I can definitely clean it up.

tasks/Node/README.md
```diff
@@ -0,0 +1,88 @@
...
+See the [RELEASENOTES.md](RELEASENOTES.md) file for details.
+
+# Pre-requisites
+
+The following are tools used to create these tasks and are recommended.
+
+* The latest version of [Visual Studio Code](https://code.visualstudio.com/).
+* The latest version of [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest).
+* The latest version of [Node.js](https://nodejs.org/en/download/).
+* [Typescript Compiler](https://www.npmjs.com/package/typescript) v2.2.0 or later.
```
mharlan: Not sure if you need to keep this line about the TS compiler in the ReadMe. It should be a dependency in your package.json and it gets invoked by the webpack ts-loader so its not used directly.

tasks/Node/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "@types/q": "^1.5.2",
+    "ncp": "^2.0.0",
+    "react-dev-utils": "^9.0.1",
+    "rimraf": "^2.6.3",
+    "ts-loader": "^6.0.4",
+    "typescript": "^3.0.3",
+    "webpack": "^4.39.0"
+  },
+  "scripts": {
+    "build-dev": "npm run copy-dev && tsc",
```
mharlan: Why is this command invoking the typsecript compiler directly? Aren't all of your TS files built by webpack now?

tasks/Node/README.md
leovms: I listed it to be in line with what is written for tasks in general. I think we may be one of the few that use Webpack to package tasks.

When the user want to compile things for testing they may run `tsc` directly. But you're correct about the `package.json`. It could be removed from the `README.md`.

tasks/Node/scripts/clean.js
```diff
@@ -0,0 +1,47 @@
...
+    if (err) {
+      console.error('Error cleaning directory: ' + destDir);
+      process.exit(1);
+    }
+
+    console.log('Cleaned ' + destDir);
+  });
+}
+
+function cleanModules(modulesDir) {
```
mharlan: Not sure why this command doesn't use rimraf. Is this leftover from when you installed node_modules within each task folder prior to the webpack bundling?

tasks/Node/package.json
leovms: When testing locally, we do need to run `tsc`. So, when setting things up for local we do use it independent of packaging. We only use Webpack for packaging.

tasks/Node/src/modules/task-utils/deployutil.ts
```diff
@@ -0,0 +1,241 @@
...
+function addParameterOverrides(parameterOverrides: string, existingParameters: DevTestLabsModels.ArmTemplateParameterProperties[]): DevTestLabsModels.ArmTemplateParameterProperties[] {
+    if (parameterOverrides == null ||
+        parameterOverrides == undefined ||
+        parameterOverrides.length == 0) {
+        return existingParameters;
+    }
+
+    let newParameter: DevTestLabsModels.ArmTemplateParameterProperties = Object.create(DevTestLabsMappers.ArmTemplateParameterProperties);
+
+    var parsedParameterOverrides = parameterOverrides.split(' ');
```
mharlan: IMO we should avoid using var. Should use either let or const.

tasks/Node/scripts/clean.js
leovms: It is meant to clean in subdirectories. I didn't know how to accomplish it with `rimraf`, I suppose. If the function can be converted, I'd be happy to. For now, it does what's needed. I call `clean` which does use `rimraf`.

tasks/Node/src/modules/task-utils/deployutil.ts
leovms: Agreed. Left over. Fixed.

tasks/Node/webpack.config.js
leovms: I'll leave it. No harm.

tasks/Node/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "tough-cookie": "^3.0.1",
+    "uuid": "^3.3.2"
+  },
+  "devDependencies": {
+    "@types/mocha": "^5.2.7",
+    "@types/ncp": "^2.0.1",
+    "@types/node": "^10.14.14",
+    "@types/q": "^1.5.2",
+    "ncp": "^2.0.0",
+    "react-dev-utils": "^9.0.1",
```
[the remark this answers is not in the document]
leovms: Thank you. Yes. Done. Testing changes don't have a negative impact.

tasks/Node/webpack.config.js
leovms: Removed.

tasks/Node/package.json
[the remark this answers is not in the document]
mharlan: The react-dev-utils is being used in your build script to print friendly webpack build errors. Can be removed it that formatting is removed from your bundler.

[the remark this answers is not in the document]
leovms: Can you elaborate on what would need to be removed from the bundler? I ran it after removing it and doing a clean fetch and it seemed to work.

tasks/Node/src/modules/task-utils/envutil.ts
```diff
@@ -0,0 +1,63 @@
...
+        const templateFileLocation = `${exportEnvTemplateLocation}/${templateFileName}`;
+        const data = response.readableStreamBody.read().toString();
+        fs.writeFileSync(templateFileLocation, data, 'utf8');
+        console.log(`Environment template has been exported to file: ${templateFileLocation}`);
+    }
+
+    console.log('Environment template has been exported successfully.');
+}
+
+export async function setOutputVariables(armClient: ResourceManagementClient, envRgId: string, template: any): Promise<any> {
```
mharlan: I've seen this in a few places, but you want Promise<void> not Promise<any> for promises that don't return anything when resolved.

tasks/Node/package.json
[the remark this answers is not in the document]
leovms: Scratch that. I thought it did work. It didn't. So, re-adding it.

tasks/Node/src/tasks/AzureDtlCreateEnvironment/task.ts
```diff
@@ -0,0 +1,148 @@
...
+
+        tl.setResult(tl.TaskResult.Succeeded, `Lab Environment '${inputData.envName}' was successfully created.`);
+    }
+    catch (error) {
+        console.debug(error);
+        tl.setResult(tl.TaskResult.Failed, error.message);
+    }
+}
+
+const args = require('minimist')(process.argv.slice(2));
```
mharlan: Saw this in a few places. Why are you using require for minimist instead of adding an import at the top of the file?

leovms: In case I wanted to remove it. I was not sure, as I only used it for testing. That's all. Is there harm doing it this way?

tasks/Node/src/modules/task-utils/envutil.ts
leovms: Done.

tasks/Node/package.json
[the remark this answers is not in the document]
mharlan: Its all the formatWebpackMessages stuff.

tasks/Node/src/tasks/AzureDtlCreateEnvironment/task.ts
mharlan: Why would you want to use multiple import types in the same file? I like consistency and this is inconsistent IMO everything should be import or require, not mixed. Since you are targeting ES6 you should use import.

mharlan: What is the issue? Problems with their default export? I know when using babel in webpack this wouldn't make any difference, Should be able to resolve commonJs export susing the typescript transpiler as well I would expect.

PR #566, opened 2019-08-23
FIX: tests not to fail if one fails and publish results

lucabol:
Running tests in parallel we had the problem that the first one failing will stop execution, so you could not tell how many failing tests you had. It was either all good or one failed.

Also, we were not publishing the test results. Now we do so that the Test pipeline tab gets populated. See the latest builds for examples.

### DIFF
--- a/samples/ClassroomLabs/Modules/Library/Tests/RunPesterTests.ps1
+++ b/samples/ClassroomLabs/Modules/Library/Tests/RunPesterTests.ps1
@@ -4,25 +4,20 @@ Param(
 
     [Parameter(Mandatory = $false, HelpMessage="If we should run the tests in 'verbose' mode for extra logging")]
     [bool] $VerboseTests = $false
-
 )
 
-# Import the module here to make sure we validate up front versions of Azure Powershell
 Import-Module $PSScriptRoot\..\Az.LabServices.psm1
 
-# We don't want to give up on the rest after a single error
-$ErrorActionPreference="Continue"
+# This allows the parallel tests not to give up on all of them when one test fails. Using just 'Continue' doesn't do it. Mystery.
+$ErrorActionPreference="SilentlyContinue"
 
-# Check if we have a newer version of Pester, if not - let's install it
 $pesterModule = Get-Module -ListAvailable | Where-Object {$_.Name -eq "Pester"} | Sort-Object -Descending Version | Select-Object -First 1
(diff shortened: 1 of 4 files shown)


samples/ClassroomLabs/Modules/Library/Tests/RunPesterTests.ps1
```diff
@@ -4,25 +4,20 @@ Param(
...
 
-# Import the module here to make sure we validate up front versions of Azure Powershell
 Import-Module $PSScriptRoot\..\Az.LabServices.psm1
 
-# We don't want to give up on the rest after a single error
-$ErrorActionPreference="Continue"
+# This allows the parallel tests not to give up on all of them when one test fails. Using just 'Continue' doesn't do it. Mystery.
+$ErrorActionPreference="SilentlyContinue"
 
-# Check if we have a newer version of Pester, if not - let's install it
```
petehauge: I looks like the existing comments were removed?  Just wondering if that was intentional?

samples/ClassroomLabs/Modules/Library/Tests/RunPesterTests.ps1
```diff
@@ -64,12 +66,5 @@ else {
         $jobs += Start-Job -Script $invokePesterScriptBlock -ArgumentList $_, $PSScriptRoot, $VerboseTests
     }
 
-    $jobs | ForEach-Object {
-        Write-Output "-----------------------------------------------------------------"
-        $result = Receive-Job -Job $_ -Wait -Verbose
-
-        if ($result.FailedCount -ne 0) {
-            Write-Error "Pester returned errors for $($result.TestResult.Describe) - $($result.TestResult.Context)"
```
petehauge: I'll check out the latest results - last time I tried this approach, even if there were test failures they weren't reported (the pipeline always showed as passed even if there were failures... )  When a test fails, does it show the pipeline as failed correctly?

samples/ClassroomLabs/Modules/Library/Tests/RunPesterTests.ps1
lucabol: Yes. I understood the logic just glancing at the code. So removed them to make the code more concise.

samples/ClassroomLabs/Modules/Library/Tests/RunPesterTests.ps1
lucabol: As discussed, it works by failing the 'publish test result' step in the pipeline.

PR #567, opened 2019-08-24
Moving DTL Tests to the new framework

lucabol:
This PR moves the DTL tests to the new framework (aka hosting in azure and reporting tests). The only thing left to do is to separate them as slow/fast.  I have done a sample run (all green) and seems to work. I have not tested a failure case, but it should logically behave as the Az.LabServices one by failing the publishing step.


petehauge:
Looks great, thanks!

PR #571, opened 2019-09-17
Setup template machine to use Hyper-V

EMaher:
Script to create virtual switch for classroom lab template machines that use Hyper-V and needs the client virtual machines to have internet connectivity.


samples/ClassroomLabs/Scripts/HyperV/SetupForNestedVirtualization.ps1
```diff
@@ -0,0 +1,288 @@
+<#
```
leovms: The README.md is recognized as a binary file. Please fix that.

Since an update is required, can you also please ensure to remove all Tab characters and that the indentation and formatting is correct?

EMaher: Both documents have been reformatted.

PR #573, opened 2019-09-25
Fixed Adding a DHCP Router to the DHCP Scope.

VaijanathB:
Fixed encoding issue in the reg key. 
Also escaped a reserve key for <Exit> as < is reserved for future use.


VaijanathB:
> Where did you get the dns server ip? The other ip addresses are hard coded because they are in the internal IP address reserved range. The dns server up is not. Is it a known azure dns server? If so we should add a comment as such.

That is AZURE DNS IP and a static IP. https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances#azure-provided-name-resolution

PR #601, opened 2020-01-02
Update Artifactfile.json

Tony75:
To allow VS2019 installation.


leovms:
@Tony75, we have not heard from you since Jan. When you are ready, feel free to resubmit.

PR #603, opened 2020-01-09
Update Az.DevTestLabs2.psm1

francedot:
Add 'storageType' and 'notes' parameter to function New-AzDtlVm


francedot:
Reopened from #604

PR #606, opened 2020-01-17
Update Az.DevTestLabs2.psm1

francedot:
Fix 'vm' variable not set in the foreach loop


petehauge:
Good catch & good fix, thanks!

PR #609, opened 2020-01-29
Enable shared Image Gallery support for creating VMs in DTL Library

petehauge:
Enable SIG support when creating VMs!  Adds a SharedImageGalleryImage parameter (multiple formats accepted) and additional parameter sets as needed.


lucabol:
Should we add a test?

petehauge:
Yes!  I'll add a test as soon as I get a chance - in the meantime, are you OK if we check this in (to unblock work with a customer)?  Thanks!

lucabol:
Done. Added #610 to track adding the test.

PR #621, opened 2020-02-26
Fix DSC parameter SharePointVersion that relied on the values of SKUs of SharePoint images

Yvand:
### Changelog

* Use hardcoded values for DSC parameter SharePointVersion as it expects "2013", "2016" or "2019", which no longer match the values used for the SKU of SharePoint images

@leovms I submitted previous PR without thoroughly testing it, sorry for that.
This time I tested a full deployment and this PR is safe to approve.


leovms:
Thank you for circling back. We rely on you to validate things, as we cannot test every scenario. So, appreciate you re-testing things once more and finding this issue.

PR #636, opened 2020-04-02
Add functions to list/start/stop student VMs

lucabol:
This introduces the basic functions for vm listing/starting/stopping when called from a student account. This is a bit on the rough side as it doesn't tie well with the rest of the library, but workeable for now. In the future we might consider making it better integrated.

### DIFF
--- a/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
+++ b/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
@@ -1144,6 +1144,140 @@ function Get-AzLabVm {
     }
     end { }
 }
+# Decided on a separate function instead of adding a switch to Get-AzLabStudentVm as the concept of Status doesn't make sense.
+# It would create a situation when you could pass Status=Stopped and -Current at the same time.
+function Get-AzLabStudentCurrentVm {
+    [CmdletBinding()]
+    param()
+    try {
+        $callerEA = $ErrorActionPreference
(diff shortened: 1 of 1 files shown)


samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
```diff
@@ -1144,6 +1144,140 @@ function Get-AzLabVm {
...
+# It would create a situation when you could pass Status=Stopped and -Current at the same time.
+function Get-AzLabStudentCurrentVm {
+    [CmdletBinding()]
+    param()
+    try {
+        $callerEA = $ErrorActionPreference
+        $ErrorActionPreference = 'Stop'
+
+
+        if($IsLinux) {
```
EMaher, line 1157: Where is $IsLinux defined?

lucabol: It should be a system variable defined by PS1 6.0+. Given that is not working for you I have changed the 'if' statement. Let's see if this one works ...

EMaher: I was testing on Mac just because it was convenient.  There is an $IsMacOs variable, too.  That's why the original code didn't work.  Sorry about that.  Maybe just return $null for Macs?  I think the system variable route is probably better as $env:os isn't guaranteed.

PR #641, opened 2020-04-15
Added Big Data Analytics class scripts

nicolehaugen:
A portion of the Big Data Analytics class type uses a PowerShell script to improve the student's experience by automatically starting Docker containers when they log in.

Note that this is the most efficient way that I found to launch Docker containers that have a dependency between them.  There are other more preferred options, such as using Docker compose.  However, this is *not* how HortonWorks has their Docker deployment scripted, so in this case it made sense just to create a PowerShell script that handles this.

The steps outlined here have been verified by walking through with the customer.


samples/ClassroomLabs/Scripts/BigDataAnalytics/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Windows Task Scheduler is used to automatically run this script when the student logs into their VM.
+
+> [!NOTE] This script assumes that you have already installed Docker and deployed Hortonworks Data Platform.  Refer to the article on the [Big Data Analytics](TODO) class type for more information.
+
+## Directions
+1. [Connect to the template machine](https://docs.microsoft.com/azure/lab-services/classroom-labs/how-to-create-manage-template#update-a-template-vm) for your class.
+1. Download the **Start-Hortonworks-DockerContainers.ps1** PowerShell script and the **Hortonworks-Scheduled-Task.xml** Windows Task Scheduler file onto the template machine through one of these ways:
+     - Clone this repository
+     - Manually download the file
```
EMaher, line 14: files

EMaher:
Changes look good to me.

PR #646, opened 2020-04-29
Updated to handle nested virtualization on Windows 10.  With the warn…

RogerBestMsft:
Script updated to handle Win 10 with caveats

### DIFF
--- a/samples/ClassroomLabs/Scripts/HyperV/README.md
+++ b/samples/ClassroomLabs/Scripts/HyperV/README.md
@@ -2,8 +2,9 @@
 This script will help prepare your template virtual machine for a classroom lab to use Hyper-V nested virtualization.
 
 ## Notes
-- This script is written for Windows Server OSes, not for Windows Client OSes.
 - Script must be executed using administrator privileges.
+- If running this script on a client O.S. like Windows 10 which does not support DHCP Server, the IP settings will need to be set within the VMs.
+    - see https://support.microsoft.com/en-us/help/15089/windows-change-tcp-ip-settings for more details on Windows 10/8/7 operating systems.
 
 # Directions
 1. Connect to template machine for your class.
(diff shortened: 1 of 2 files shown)


samples/ClassroomLabs/Scripts/HyperV/SetupForNestedVirtualization.ps1
```diff
@@ -283,7 +320,14 @@ try {
         Write-Host "Unable to set WinNat service to Automatic.  Hyper-V virtual machines will not have internet connectivity when service is not running." -ForegroundColor Yellow
     }  
 
-    # Tell the user script is done.
+    # Tell the user that VMs will need to set their own IPs on Windows 10
+    # Link for Windows machines https://support.microsoft.com/en-us/help/15089/windows-change-tcp-ip-settings
+    if ($(Get-RunningServerOperatingSystem ) -eq $false) {
+        Write-Host -Object "DHCP Server is not supported on Windows 10. `
+        IP configuration will need to be done manually from within the VM itself `
```
EMaher, line 327: What do you think about changing the warning to something like
"IP addresses for each Hyper-V VM.  Address must be 192.168.0.4 and 192.168.0.254 with a subnet mask of 255.255.255.0.  To use the Azure Dns use 168.63.129.16"

This would avoid people have to grep through the script to find the values.

samples/ClassroomLabs/Scripts/HyperV/README.md
```diff
@@ -2,8 +2,9 @@
 This script will help prepare your template virtual machine for a classroom lab to use Hyper-V nested virtualization.
 
 ## Notes
-- This script is written for Windows Server OSes, not for Windows Client OSes.
 - Script must be executed using administrator privileges.
+- If running this script on a client O.S. like Windows 10 which does not support DHCP Server, the IP settings will need to be set within the VMs.
```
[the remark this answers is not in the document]
EMaher: Nevermind, I see that dns also needs to be set.

samples/ClassroomLabs/Scripts/HyperV/SetupForNestedVirtualization.ps1
RogerBestMsft: Agreed, I've made the change.

RogerBestMsft:
Issues found, will open new pr with fixes

PR #689, opened 2020-08-18
Revert SP and SQL to old VM sizes

Yvand:
Since today, all [Edsv4-series](https://docs.microsoft.com/en-us/azure/virtual-machines/edv4-edsv4-series) VMs fail to start in region "East US", which impacts SharePoint template since it's used for SP and SQL VMs.
So I revert SP and SQL to their old VM size as a workaround.


Yvand:
This PR only intends to fix an issue with availability Edsv4-series VMs in "East US" (that surfaced just today).
My guess is that it's pure bad luck that we face this problem, and it's totally unrelated to other issues in the past few weeks.
I prefer to revert to old sizes until Edsv4-series works again, as internally we use "East US" region a lot.

PR #692, opened 2020-09-05
Update VM Sizes and Container dependency

cliveg:
Including v4 VM SKU's and Windows 'container' feature, and fix hyphen bug.


leovms (approved):

PR #723, opened 2020-12-15
Docker script fails to execute and hangs very long time

luszanko:
The docker script fails each time I want to install it on Win Srv 2016. After I add the tls12 setting the installation completes successfully.


leovms (approved):
Thank you for your contribution.

PR #731, opened 2021-02-02
Add Teams support to Lab Services PowerShell Library

petehauge:
## Details
In the Azure Lab Services Rest APIs we can now create labs that are connected with MS Teams via using the aadGroupId property. This change includes both updates to the PowerShell Module for Lab Services and the Lab Creator tool.

There are a set of other changes included in this pull request:
* Enable passing "aadGroupId" in the library & lab creator script to connect a lab to teams
* Update the example CSV file with aadGroupId as a column
* Update our ThreadJob module install so that it installs per-user instead (so we don't need an admin powershell window to run it)
* Add in "LabOwnerEmail" in the CSV file and in the Lab Creator script.  This way we can have a separate set of users who are 'Lab Owners' for bulk creation of labs
* Add in validation for the Lab Name up front - if there's an invalid character we want to fail fast
* Write a warning if the user of the Lab Creator tool doesn't update the username/password in the sample CSV file
* Update the conversion code for boolean/linked values to account for missing, empty & valid values.  (we would fail for missing before)


samples/ClassroomLabs/Modules/Library/Tools/LabCreator.ps1
```diff
@@ -10,7 +10,7 @@ param(
 )
 
 Import-Module ../Az.LabServices.psm1 -Force
-Install-Module -Name ThreadJob -Force
+Install-Module -Name ThreadJob -Scope CurrentUser -Force
```
petehauge, line 13: Switching to installing only per user, so we can run this without needing admin permissions on the powershell window

samples/ClassroomLabs/Modules/Library/Tools/LabCreator.ps1
```diff
@@ -267,20 +271,42 @@ $labs = Import-Csv -Path $CsvConfigFile
 Write-Verbose ($labs | Format-Table | Out-String)
 
 $labs | ForEach-Object {
-    if ($_.Emails) {
+
+    # Validate that the name is good, before we start creating labs
+    if (-not ($_.LabName -match "^[a-zA-Z0-9_, '`"!|-]*$")) {
+        Write-Error "Lab Name '$($_.LabName)' can't contain special characters..."
```
petehauge, line 277: Adding lab name validation up front to catch any invalid names so we can fail early (rather than fail later)

samples/ClassroomLabs/Modules/Library/Tools/LabCreator.ps1
```diff
@@ -267,20 +271,42 @@ $labs = Import-Csv -Path $CsvConfigFile
...
+    }
+
+    if ((Get-Member -InputObject $_ -Name 'Emails') -and ($_.Emails)) {
         $_.Emails = ($_.Emails.Split(';')).Trim()
     }
 
     if (Get-Member -InputObject $_ -Name 'GpuDriverEnabled') {
-        $_.GpuDriverEnabled = [System.Convert]::ToBoolean($_.GpuDriverEnabled)
+        if ($_.GpuDriverEnabled) {
```
petehauge, line 285: Fixing the parameter validation - we need to handle the case where it's missing, present but blank, and present with a value...    Same fix for LinuxRdp.

samples/ClassroomLabs/Modules/Library/Tools/LabCreator.ps1
```diff
@@ -267,20 +271,42 @@ $labs = Import-Csv -Path $CsvConfigFile
...
+            $_.LinuxRdp = $false
+        }
+    }
+    else {
+        Add-Member -InputObject $_ -MemberType NoteProperty -Name "LinuxRdp" -Value $false
     }
 
     $_.SharedPassword = [System.Convert]::ToBoolean($_.SharedPassword)
-    if ($_.Schedules) {
+    if ((Get-Member -InputObject $_ -Name 'Schedules') -and ($_.Schedules)) {
```
petehauge, line 309: Fixing to make sure the script can run if the Schedules entry is missing (not present in the CSV file)

samples/ClassroomLabs/Modules/Library/Tools/LabCreator.ps1
lucabol: Should this be in the library instead?

petehauge: Thanks!  We agreed offline to keep it here for now - we can always change it later based on feedback...

samples/ClassroomLabs/Modules/Library/Tools/LabCreator.ps1
```diff
@@ -125,20 +133,33 @@ $init = {
...
             Write-Host "Linux $LinuxRdp***"
     
             $lab = $la `
             | New-AzLab -LabName $LabName -Image $img -Size $Size -UserName $UserName -Password $Password -LinuxRdpEnabled:$LinuxRdp -InstallGpuDriverEnabled:$GpuDriverEnabled -UsageQuotaInHours $UsageQuota `
-                -idleGracePeriod $idleGracePeriod -idleOsGracePeriod $idleOsGracePeriod -idleNoConnectGracePeriod $idleNoConnectGracePeriod `
+                -idleGracePeriod $idleGracePeriod -idleOsGracePeriod $idleOsGracePeriod -idleNoConnectGracePeriod $idleNoConnectGracePeriod -AadGroupId $AadGroupId `
             | Publish-AzLab `
             | Set-AzLab -MaxUsers $MaxUsers -UserAccessMode $UsageMode -SharedPasswordEnabled:$SharedPassword
 
+            # If we have any lab owner emails, we need to assign the RBAC permission
```
EMaher, line 146: Nice!

samples/ClassroomLabs/Modules/Library/Tools/hogwarts.csv
```diff
@@ -1,4 +1,4 @@
...
-hogwarts-rg2,westeurope,Wizardy,Transfiguration,CentOS-Based*,2,20,Restricted,false,Basic,"Transfiguration","Who are you?",test0000,Test00000000,true,"lucabol@microsoft.com;lucabol@microsoft.com", 7,8,9,"Transfiguration"
-hogwarts-rg2,westeurope,Wizardy,Charms,CentOS-Based*,2,20,Restricted,false,Basic,"Charms","Encanted",test0000,Test00000000,true,"lucabol@microsoft.com;lucabol@microsoft.com",15,0,10, "Charms", charms
\ No newline at end of file
+ResourceGroupName,Location,LabAccountName,LabName,ImageName,AadGroupId,MaxUsers,UsageQuota,UsageMode,SharedPassword,Size,Title,Descr,UserName,Password,LinuxRdp, Emails,LabOwnerEmails, idleGracePeriod, idleOsGracePeriod, idleNoConnectGracePeriod, Invitation, Schedules
+hogwarts-rg2,westeurope,Muggle,"History of Magic",CentOS-Based*,00000000-0000-0000-0000-000000000000,2,20,Restricted,FALSE,Basic,"History of Magic","Those who don't know history ...",test0000,Test00000000,TRUE,"lucabol@microsoft.com;lucabol@microsoft.com","Peter.Hauge@microsoft.com", 7,8,9,"History of Magic"
```
EMaher, line 2: @petehauge You changed all the booleans to upper case.  Is there a reason for this?  Just want to make sure we aren't breaking anyone already using the tool that uses lower case true/false.

EMaher (approved):
LGTM

samples/ClassroomLabs/Modules/Library/Tools/hogwarts.csv
petehauge: This happened when I opened the CSV file in excel & made changes...  It's not required for the scripts, the casing for PowerShell doesn't matter.  If you want me to switch it back, just let me know - but I figured it would be common that customers open it in excel.  Thanks!

lucabol (approved):

samples/ClassroomLabs/Modules/Library/Tools/hogwarts.csv
EMaher: No, that's fine.  Just checking.

PR #740, opened 2021-02-11
Updates to the LabCreator.ps1 bulk lab creation tool

petehauge:
# Description
Some additional updates included with this pull request:
* Add timings to the script for overall execution and individual lab creation
* Enable reporting output to the user over time (not all at once at the end)
* Enable continuing working on the rest of the labs even if a single one fails (don't stop on a single lab failure)
* Clean up memory for jobs along the way for labs (using Remove-Job)
* Enable creating labs without a template VM
* Add some color to the output to make it pretty
* Only install the ThreadJob module if it's not already available
* Catch a case where LabOwner emails field doesn't exist at all (or is null)


RogerBestMsft (approved):
Add explicit call out in the readme for the parameter change.  Rest looks good.

lucabol (approved):
Looks good to me.

PR #765, opened 2021-05-05
Added GPU template VM restart and misc fixes

nicolehaugen:
When GPU drivers are installed, I added an automatic restart of the template VM.  For the GPU drivers to properly install, the machine must be restarted afterwards.  The product should actually ensure that the template VM is restarted - however, it's rare that users run into this in the UI because they typically start the template VM at least once before publishing.

While adding this, I also ran across a few other bugs in the script which I fixed in this PR:
1.) The LinuxRdp property wasn't working because we had a mismatch in naming - we expected the .csv column to be named "LinuxRdp" but the New-AzLab function expected it to be "LinuxRdpEnabled".
2.) Same issue as mentioned in the above bullet, also existed for the GpuDriverEnabled properly.
3.) The SharedPassword property didn't work in the case that you had an existing lab that it was attempting to call Set-AzLab with.  This is because of an inconsistency issue with SharePassword expecting Enabled\Disabled instead of true\false.  Also, there is a naming consistency issue with this property.  Refer to the comments in the code for more info.
4.) My .csv file structure failed because it didn't included several columns which shouldn't be required.  I changed the import csv function to default these unrequired columns to empty values where needed.

### DIFF
--- a/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
+++ b/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
@@ -885,7 +885,7 @@ function New-AzLab {
 
                     # The template VM needs to be restarted for the GPU drivers to finish installing properly
                     # This will eventually be done within the product, but for now, we need to do this explicitly
-                    Write-Verbose "Start template VM"
+                    Write-Verbose "Start template VM after installing GPU drivers"
                     Get-AzLabTemplateVm $lab | Start-AzLabTemplateVm
                     Get-AzLabTemplateVm $lab | Stop-AzLabTemplateVm
                     Write-Verbose "Stopped template VM"
--- a/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
+++ b/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
@@ -917,9 +917,9 @@ function Set-AzLab {
         [string]
         $UserAccessMode,
 
-        [parameter(Mandatory = $false, ValueFromPipelineByPropertyName = $true)]       
+        [parameter(Mandatory = $false, ValueFromPipelineByPropertyName = $true)]
         [ValidateSet('Enabled', 'Disabled')]
-         [string]
+        [string]
         $SharedPasswordEnabled,
 
         [parameter(Mandatory = $false, ValueFromPipelineByPropertyName = $true, HelpMessage = "Quota of hours x users (defaults to 40)")]
--- a/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
+++ b/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
@@ -776,7 +776,7 @@ function New-AzLab {
             
         [parameter(mandatory = $false, ValueFromPipelineByPropertyName = $true)]
         [switch]
-        $InstallGpuDriverEnabled = $false,
+        $GpuDriverEnabled = $false,
 
         [parameter(Mandatory = $true, ValueFromPipelineByPropertyName = $true, HelpMessage = "User name if shared password is enabled")]
         [string]
@@ -788,7 +788,7 @@ function New-AzLab {
 
         [parameter(mandatory = $false, ValueFromPipelineByPropertyName = $true)]
         [switch]
-        $LinuxRdpEnabled = $false,
+        $LinuxRdp = $false,
 
         [parameter(Mandatory = $false, ValueFromPipelineByPropertyName = $true, HelpMessage = "Quota of hours x users (defaults to 40)")]
         [int]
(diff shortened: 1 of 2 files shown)


samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
```diff
@@ -879,7 +879,17 @@ function New-AzLab {
...
                 $lab = WaitProvisioning -uri $labUri -delaySec 60 -retryCount 120
                 #Wait for template to be provisioned.  Even labs without a template will return a environmentsetting object.
-                $defaultEnvironmentSetting = WaitProvisioning -uri $environmentSettingUri -delaySec 60 -retryCount 120 
+                $defaultEnvironmentSetting = WaitProvisioning -uri $environmentSettingUri -delaySec 60 -retryCount 120
+                
+                if ($GpuDriverEnabled) {
+
+                    # The template VM needs to be restarted for the GPU drivers to finish installing properly
+                    # This will eventually be done within the product, but for now, we need to do this explicitly
+                    Write-Verbose "Start template VM"
```
lucabol, line 888: I would specify 'Start Template VM after installing GPU drivers' otherwise people might wander why we are doing this. You can then remove the comment.

samples/ClassroomLabs/Modules/Library/BulkOperations/Az.LabServices.BulkOperations.psm1
```diff
@@ -84,10 +84,36 @@ function Import-LabsCsv {
...
+            Add-Member -InputObject $_ -MemberType NoteProperty -Name "LabOwnerEmails" -Value @() 
+        }
+
+        # TODO: There are two odd things about this code:
+        #   1.) The name of this property on the input object is "SharedPassword", but the the New-AzLab function expects the parameter name to be "SharedPasswordEnabled".
+        #   2.) The Set-AzLab function expects this property to have Enabled\Disabled values, but New-AzLab expects true\false.
+        # In the future, we need to resolve these inconsistencies. It works right now because:
+        #   1.) New-AzLabsBulk calls New-AzLab to create a new lab.  When New-AzLab is called, it ignores this "SharedPassword" property value and defaults it to False because New-AzLab is expecting the property to be named "SharedPasswordEnabled".
+        #   2.) New-AzLabsBulk then calls Set-AzLab.  When Set-AzLab is called, the value of this "SharedPassword" property is explicitly passed in as the "SharedPasswordEnabled" parameter.
+        if (Get-Member -InputObject $_ -Name 'SharedPassword') {
```
lucabol, line 106: Great comment!

lucabol (approved):

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
```diff
@@ -829,13 +829,13 @@ function New-AzLab {
                 $environmentSettingUri = $labUri + "/environmentsettings/default"
                 $sharedPassword = if ($SharedPasswordEnabled) { "Enabled" } else { "Disabled" }
                 $imageType = if ($image.id -match '/galleryimages/') { 'galleryImageResourceId' } else { 'sharedImageResourceId' }
-                if ($LinuxRdpEnabled) {$linuxRdpState = 'Enabled'} else { $linuxRdpState = 'Disabled' }
-                if ($InstallGpuDriverEnabled) {$gpuDriverState = 'Enabled'} else { $gpuDriverState = 'Disabled' }
+                if ($LinuxRdp) {$linuxRdpState = 'Enabled'} else { $linuxRdpState = 'Disabled' }
+                if ($GpuDriverEnabled) {$gpuDriverState = 'Enabled'} else { $gpuDriverState = 'Disabled' }
```
petehauge, line 833: I wonder, since you're renaming anyway - if we should follow the same pattern for LinuxRdp & GpuDriver...  ?   They are used as a switch parameter, so it's implied that passing the switch means "Enabled".

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
```diff
@@ -829,13 +829,13 @@ function New-AzLab {
...
-                if ($LinuxRdpEnabled) {$linuxRdpState = 'Enabled'} else { $linuxRdpState = 'Disabled' }
-                if ($InstallGpuDriverEnabled) {$gpuDriverState = 'Enabled'} else { $gpuDriverState = 'Disabled' }
+                if ($LinuxRdp) {$linuxRdpState = 'Enabled'} else { $linuxRdpState = 'Disabled' }
+                if ($GpuDriverEnabled) {$gpuDriverState = 'Enabled'} else { $gpuDriverState = 'Disabled' }
                 if ($idleGracePeriod -eq 0) {$idleShutdownMode = "None"} else {$idleShutdownMode = "OnDisconnect"}
                 if ($idleOsGracePeriod -eq 0) {$enableDisconnectOnIdle = "Disabled"} else {$enableDisconnectOnIdle = "Enabled"}
                 if ($idleNoConnectGracePeriod -eq 0) {$enableNoConnectShutdown = "Disabled"} else {$enableNoConnectShutdown = "Enabled"}
 
-                if ($LinuxRdpEnabled) {
+                if ($LinuxRdp) {
```
petehauge, line 838: This code looks wrong - but it's probably fine if we don't address it right now (maybe log a task for it).  It looks like we are deciding to set the Idle settings IF the LinuxRDP property is passed.  I don't think that's right because some Linux operating systems do support idle settings.  We should either be asking the user if it's a LinuxOS (and making both the RDP & idle settings decision based on that), or have 2 properties...  We can fix this later though, definitely don't need to hold up the PR for it.

samples/ClassroomLabs/Modules/Library/BulkOperations/Az.LabServices.BulkOperations.psm1
```diff
@@ -84,10 +84,36 @@ function Import-LabsCsv {
...
+            }
+            else {
+                $_.SharedPassword = 'Disabled'
+            }
+        }
+        else {
+            Add-Member -InputObject $_ -MemberType NoteProperty -Name "SharedPassword" -Value 'Disabled'
+        }
 
         if (Get-Member -InputObject $_ -Name 'GpuDriverEnabled') {
```
petehauge, line 118: Something we should include here - if the user has No Template specified AND GpuDriverEnabled, we should probably fail validation right?  Because we need a template for GPU drivers today, so we can fail up front and tell the user they need to pick one or the other...

petehauge (approved):

EMaher (approved):

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
nicolehaugen: I changed the names of these to match the names of the columns currently expected\enforced via the Az.LabServices.BulkOperatations.psm1 script - if we change these, then we'll need to change the names of the corresponding columns passed in via the .csv file.  This feels like more of a breaking change which I don't think is important right now.  We should improve and address this with vNext.  I can log a bug to track this as an issue.

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
nicolehaugen: Ah yes - idle setting support for some Linux OS's wasn't added until fairly recently, so this code never got updated to reflect that. I'll log a task for this as well as part of the vNext rework.

samples/ClassroomLabs/Modules/Library/BulkOperations/Az.LabServices.BulkOperations.psm1
nicolehaugen: I thought about it, but decided to hold off because in the email thread where we're discussing shortcomings of the template-less scenario, they mentioned considering ways to allow the GPU drivers to be installed even in the template-less scenario.  I will log a task to consider this for vNext depending on the behavior they decide.

PR #775, opened 2021-05-27
Update azuredeploy-function.json

jrudley:
Updated storage account api in order to correctly lookup azure gov cloud endpoint.


leovms (approved):

PR #782, opened 2021-06-22
Update template sharepoint full

Yvand:
## Changelog

* Reduce deployment time by skipping creation of developer site /sites/dev, not so useful
* Reduce deployment time by enabling the distributed cache service during the SharePoint farm creation (in SP VM only)
* Reduce deployment time by running script UpdateGPOToTrustRootCACert only if necessary
* Install Visual Studio Code in SP and FE VMs
* Create modern team sites instead of classic team sites in SharePoint 2019
* Return various information as output of the template deployment
* Update TLS 1.2 settings in SP and FE VMs
* Enable file sharing (on Domain network profile) also on SQL VM (it is already enabled on SP and FE VMs)
* Update DSC module SharePointDSC from 4.5.1 to 4.7, removed the now useless dependency on ReverseDSC and manually added the changes in PR #1325
* Update DSC module xDnsServer from 1.16.0 to 2.0
* Set local admin name on VM SQL/SP/FE with a unique string, to avoid using the local admin instead of the domain admin


Yvand:
@leovms this PR is ready on my side. Is there anything blocking?

leovms (approved):

PR #785, opened 2021-07-08
Update artifact.ps1

jrudley:
I am testing on Windows 10 and the command ```Get-WUInstall -IgnoreReboot -AcceptAll``` just lists the updates, but doesn't download or install them. I updated the command to ```Get-WUInstall -IgnoreUserInput -Acceptall -Download -Install -Verbose -IgnoreReboot``` which will download and install.


leovms (approved):

PR #788, opened 2021-07-12
create Windows Defender ATP Local Onboradng Script

pksai:
This scrip will run the defender


pksai:
script will run the defender

PR #810, opened 2021-10-05
Installing updates for all Microsoft products

jonasblunck:
Update windows-install-windows-updates artifact to enable installing updates for all Microsoft products, not only Windows.


leovms (approved):
Thank you for the enhancement.

PR #827, opened 2022-02-04
windows-visualstudio - added VS2019 pro, ent and VS2022 pro, ent

mhackermsft:
Adds the ability to install the professional or enterprise version of Visual Studio 2019 or 2022.

To test, just add the artifact to a VM in DevTest Labs and choose the version and edition of Visual Studio to install.  After provisioning you will see the correct version of Visual Studio installed upon logging into the VM.

Vs2022ent

vs2022pro

vs2019ent

vs2019pro


leovms (approved):

PR #831, opened 2022-03-23
Fix authentication in Az.LabServices module to work in Azure Automation

petehauge:
Azure Automation now works with Managed Identities (this is the default when creating a new account) instead of the Run As accounts that existed before.  We needed some changes to our module to work in Azure Automation & MSIs correctly!

### DIFF
--- a/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
+++ b/samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
@@ -5,9 +5,6 @@
 if ("AzureAutomation/" -ne $env:AZUREPS_HOST_ENVIRONMENT) {
     # We are using strict mode for added safety
     Set-StrictMode -Version Latest
-
-    # We require a relatively new version of Powershell
-    #requires -Version 3.0
 }
 
 # To understand the code below read here: https://docs.microsoft.com/en-us/powershell/azure/migrate-from-azurerm-to-az?view=azps-2.1.0
@@ -214,22 +211,11 @@ function ConvertFrom-ISO8601Duration {
 
 function Get-AzureRmCachedAccessToken() {
     $ErrorActionPreference = 'Stop'
-    Set-StrictMode -Off
+    Set-StrictMode -Off 
 
     if ("AzureAutomation/" -eq $env:AZUREPS_HOST_ENVIRONMENT) {
         Write-Verbose "Running in Azure Automation environment..."
-        # We are running in Azure Automation - so need to get the token another way
-        # First, ensure that we're logged in
-        
-        $url = $env:IDENTITY_ENDPOINT
-        $headers = New-Object "System.Collections.Generic.Dictionary[[String],[String]]" 
-        $headers.Add("X-IDENTITY-HEADER", $env:IDENTITY_HEADER) 
-        $headers.Add("Metadata", "True") 
-        $body = @{resource='https://management.azure.com/' } 
-        $accessToken = Invoke-RestMethod $url -Method 'POST' -Headers $headers -ContentType 'application/x-www-form-urlencoded' -Body $body
-        write-Verbose "Access token object:"
-        $accessToken | ConvertTo-Json -depth 10 | Out-String | Write-Verbose 
-        return $accessToken.access_token
+        return (Get-AzAccessToken).Token
     }
     else {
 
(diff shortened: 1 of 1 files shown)


samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
```diff
@@ -1,11 +1,14 @@
...
+# Only set these if not running under Azure Automation
+if ("AzureAutomation/" -ne $env:AZUREPS_HOST_ENVIRONMENT) {
+    # We are using strict mode for added safety
+    Set-StrictMode -Version Latest
 
-# We require a relatively new version of Powershell
-#requires -Version 3.0
+    # We require a relatively new version of Powershell
+    #requires -Version 3.0
+}
```
petehauge, line 11: The commandlet and directive seem to work differently in Azure Automation than in regular PowerShell.  The only way I could make it work is excluding these lines if we're running in an Azure Automation environment...

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
```diff
@@ -213,32 +216,50 @@ function Get-AzureRmCachedAccessToken() {
     $ErrorActionPreference = 'Stop'
     Set-StrictMode -Off
 
-    $azureRmProfile = [Microsoft.Azure.Commands.Common.Authentication.Abstractions.AzureRmProfileProvider]::Instance.Profile
-    if (-not $azureRmProfile.Accounts.Count) {
-        Write-Error "Ensure you have logged in before calling this function."
+    if ("AzureAutomation/" -eq $env:AZUREPS_HOST_ENVIRONMENT) {
```
petehauge, line 219: When running in Azure Automation environment, we need to get the token another way (doing a direct Rest API call based on a localhost identity endpoint).  This is based on Automation accounts that use Run As, Service Principal, User-provided MSI or System-provided MSI.

RogerBestMsft (approved):

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
EMaher: @petehauge - Code itself for calling the identity endpoint looks good.

However, before we add this, did you test call `Connect-AzAccount -Identity`?  It's the supported cmdlet to connect using an MSI.  See https://docs.microsoft.com/en-us/powershell/module/az.accounts/Connect-AzAccount?view=azps-7.3.2#example-5--connect-using-a-managed-service-identity.   

If running, `Connect-AzAccount -Identity` doesn't work, then code looks good to me.

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
EMaher: Weird.  Honestly, do we even need this anymore?

EMaher (approved):

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
petehauge: I don't think so - I haven't heard of any customers on super old versions of PowerShell any longer...  I'm fine removing it!

samples/ClassroomLabs/Modules/Library/Az.LabServices.psm1
petehauge: Yes-  I agree that's the way to login - the user would need to call Connect-AzAccount -Identity in their runbook (I tested that this works), but in the library we specifically need the access token which isn't available after calling the commandlet to login with an MSI...  I'll try again using Connect-AzAccount and see if the access token is populated and let you know!
https://docs.microsoft.com/en-us/azure/automation/enable-managed-identity-for-automation

EMaher: Sounds good.  I just remember having a similar issue with the Azure PowerShell functions and that call fixed the issue.  

If that does work, I wonder what the best way to educate people would be.

petehauge: @EMaher  - good call on this, I tried with simplified code relying on Connect-AzAccount -Identity and it all worked as expected!  I've simplified the code and will get it checked in shortly after some additional testing.

PR #835, opened 2022-04-26
Update readmes to include April 2022 Update

RogerBestMsft:
Update readmes to point to new repo and gallery.


EMaher (approved):
Only comment -- Should we use a block comment (>) to call more attention to the note?

PR #842, opened 2022-06-24
fix credentials issue with dsc extension

Yvand:
## CHANGELOG

* Fix the credentials decryption issue in DSC extension when using latest version of Windows Server images.

### DIFF
--- a/Environments/SharePoint-AllVersions/CHANGELOG.md
+++ b/Environments/SharePoint-AllVersions/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change log for Azure DevTest Labs template SharePoint-AllVersions
 
+## Enhancements & bug-fixes - Published in June 24, 2022
+
+* Fix the credentials decryption issue in DSC extension when using latest version of Windows Server images.
+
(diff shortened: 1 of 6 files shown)


Yvand:
This PR is ready for review. It mitigates a new issue which seems to be caused by a [security update on Windows](https://www.reddit.com/r/PowerShell/comments/vgi99r/credential_with_dsc_issue_after_windows_update/) and affects templates which use the Azure DSC extension to pass credentials (so probably many).
As a quick workaround, I hardcode a previous version of the Windows Server images which do not cause this issue (instead of using latest).

Yvand:
@leovms is this PR in your radar?

Yvand:
The regression is now fixed by the latest version of Windows images, which contain [this update](https://support.microsoft.com/en-us/topic/july-12-2022-kb5015811-os-build-17763-3165-f3bdb13c-d767-47dc-a077-0ea0e9421a96):
`"Addresses an issue that redirects the PowerShell command output so that transcript logs do not contain any content. Transcript logs might contain decrypted passwords if you turn PowerShell logging on. Consequently, the transcript logs lose the decrypted passwords."`

Closing this PR

PR #868, opened 2023-05-03
Fix ServiceFabric-LabCluster template

Kencdk:
The template for creating a service fabric cluster using DTL, is referencing two QuickStart arm templates. Since our templates last update, both QuickStart templates have added new required parameters, causing the template to fail.

Add the three new parameters, required when deploying a Windows cluster.
Pass in the resource group location as storage location, when deploying a Linux cluster.

### DIFF
--- a/Environments/ServiceFabric-LabCluster/azuredeploy.json
+++ b/Environments/ServiceFabric-LabCluster/azuredeploy.json
@@ -76,7 +76,7 @@
                 "description": "Client id of cluster application. Required for Windows clusters."
             }
         },
-        "clientapplication": {
+        "clientApplication": {
             "type": "string",
             "defaultValue": "",
             "metadata": {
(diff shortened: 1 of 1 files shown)


Environments/ServiceFabric-LabCluster/azuredeploy.json
```diff
@@ -61,6 +61,27 @@
...
+            }
+        },
+        "clusterApplication": {
+            "type": "string",
+            "defaultValue": "",
+            "metadata": {
+                "description": "Client id of cluster application. Required for Windows clusters."
+            }
+        },
+        "clientapplication": {
```
joerage, line 79: casing? `clientApplication`

joerage: Same below

AaronPatterson:
Should we do something to prevent this from happening again?  Do we expect the quick start arm template to change again?

Kencdk:
> Should we do something to prevent this from happening again? Do we expect the quick start arm template to change again?

Ideally, for production workloads, a customer would copy in the bits they need - thus not using these samples but copying in the root arm templates and modify those per their needs.

It's hard to predict when/if the ServiceFabric team will update their samples - these are based on quickstart guides after all and subject to change whenever the team feels like it.

Issue #871, opened 2023-05-23
Unclaim VM function executing "claim" action


thophan-microsoft:
In function [Invoke-AzDtlVmUnclaim:1076](https://github.com/Azure/azure-devtestlab/blob/bff4aa88281b613f5e149fda04f6dca765ff81c5/samples/DevTestLabs/Modules/Library/Az.DevTestLabs2.psm1#LL1076C1-L1077C1), the command is executing a "claim" action on the VM when it should be "unclaim"

PR #872, opened 2023-05-24
Bug: Fix action to unclaim VM

thophan-microsoft:


thophan-microsoft:
https://github.com/Azure/azure-devtestlab/issues/871

PR #901, opened 2024-10-01
Setting public registry for npm

usanth:
Changed from using private registry to public registry

### DIFF
--- a/tasks/Node/package-lock.json
+++ b/tasks/Node/package-lock.json
@@ -39,28 +39,26 @@
       }
     },
     "node_modules/@azure/abort-controller": {
-      "version": "1.0.4",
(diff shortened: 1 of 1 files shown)


jfullerton44:
Is there no package.json change here, I just see the generated file changes

usanth:
@jfullerton44 the package.json wasn't referencing the private feed from my understanding, it was just outlining the dependencies. I just installed those using the public feed rather than the private one, which is what generated the new package-lock.json

jfullerton44:
If someone has to regenerate the package-lock file, how will it know which feed to use?

jasongin (approved):

jasongin:
> If someone has to regenerate the package-lock file, how will it know which feed to use?

It will use the public feed by default since there is no `.npmrc` file in the repo. I think someone had set up a temporary `.npmrc` file previously.

